PR #4, opened 2021-03-30
macos: interface in install

mrdomino:
Looks like `fetch` is now `up`. Also, I needed to put an interface on the command line for innernet to work.

This is maybe not exactly the correct fix — at minimum the docs should be updated to say “kick the innernet launchd unit after you set up your interface”. Or perhaps `up` should listen on all interfaces on `/etc/innernet` or something.

### DIFF
--- a/macos/install.sh
+++ b/macos/install.sh
@@ -25,6 +25,9 @@ if ! which wg > /dev/null; then
   cmd brew install wireguard-tools
 fi
 
+read -p "interface to listen on: "
+interface=${REPLY}
+
 info "installing launch daemon for innernet daemon script."
 echo "\
 <?xml version=\"1.0\" encoding=\"UTF-8\"?>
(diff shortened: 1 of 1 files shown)


mcginty:
Thanks for the PR! I think it might make sense for now to start with adding a flag in `innernet up` to bring up all interfaces, and then we can modify this launchd file to just use that flag.

In the slightly longer term, I think it makes sense to add support in the `innernet install` subcommand to write and enable launch daemons or systemd units per interface to make installation as straightforward as possible. Thoughts?

mrdomino:
Generally agree with all of that.

Seems easy enough to accomplish that with systemd, and probably the right way to do it. I don't know enough about launchd to know if there's a good way to do that, but I suppose in the worst case you can just name the units different things depending on interface.

mcginty:
@mrdomino sorry for this stalling - I think I'm going to go ahead and remove this install script all together and write some simple homebrew/macports installers instead, which should be more useful to all macOS users going forward.

PR #21, opened 2021-03-31
Fix macos ip/route setup for ipv6 nets

aeber:
When using innernet with an IPv6 network as internal network, installation and running `up` fails on MacOS. 

I believe this is due to usage of `inet` instead of `inet6` and a slightly different syntax when assigning the address to the interface.


mcginty:
Thanks! Nice catch. I'll put this in the upcoming patch release.

PR #29, opened 2021-04-03
Build rpms with github actions

mk0x9:
I've added scripts & adjusted github actions pipeline to provide rpm-based packages. Packages are built inside dockerized base distribution environments due to dynamic linking with sqlite3, otherwise it would be possible to use the same static binary everywhere.

Here is result of running updated release action: https://github.com/mk0x9/innernet/actions/runs/714004665


mcginty:
Thanks @mk0x9! I see you're building in Fedora 32&33, as well as CentOS 8. Are RPMs typically incompatible across distros? I'm a little concerned about maintainability.

mk0x9:
I'm not an expert, should be compatible, there is only a dependency on glibc & sqlite ABI. Also I'm not an expert, but maybe there are might be an issues (or not) due to use of LTO. Also I've just checked vscode repo, they're building package only for 1 distro for every arch, I would assume it's probably more than enough

Kerwood:
It sure would be great with a rpm package. Can't wait for this PR to be merged :100:

mcginty:
Sorry for the delay here. This seems reasonable to merge, but I'll be relying on others to report problems with the RPMs. Maybe we can later add more docker tests eventually to check that the distro packages can install successfully or something.

PR #31, opened 2021-04-04
systemd: Add file innernet.target

mnhauke:
Add file innernet.target, which allows you to stop or restart all client
instances.

### DIFF
--- a/client/innernet.target
+++ b/client/innernet.target
@@ -1,0 +1,2 @@
+[Unit]
+Description=Target to allow restarting and stopping of all parts of innernet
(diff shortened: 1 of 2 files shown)


mcginty:
Nice!

mnhauke:
The [Install] section is not interpreted by systemd(1) during runtime; it is used by the `enable` and `disable` commands of the systemctl(1) tool during installation of a unit.
https://www.freedesktop.org/software/systemd/man/systemd.unit.html

I'm not aware of a way to start all instantiated services of a service at once.

mcginty:
Cool, thanks for checking, and thanks for the PR. Seems totally fine to merge this as-is.

Issue #26, opened 2021-03-31
Add more protections against OS configurations that allow IP spoofing


Spindel:
In the innernet blog-post there are some claims about authenticated and trust in IP-addresses. 

However, innernet does not actually live up to these claims.

Test methodology: 

With three devices:
- one innernet server (that lives on the internet)  (10.42.0.0/16)
- one innernet client  ( NAT-ed, on fex. a WiFi)  (10.42.0.1/24)
- One rogue actor ( NAT-ed, on the same net as the client)

Assuming that the client LAN is 192.168.0.1/24.   
Assuming that the innernet server is set up with the default cidr of 10.42.0.0/16,  we hook up the client and enable the server.

Next, we put a super-seekrit service on the client IP of 10.42.0.1 ( telnet, with admin/admin as password!) 

On the "rogue" actor, we then do the following (assuming that eth0 is the ip address):

```
ip address add 10.42.0.8 dev eth0
ip route add 10.42.0.0/16 dev eth0
```

Now the "rogue" client can freely access services that are supposed to only be accessible to internal clients on the innernet ip.  

# Problem

Linux by default responds to traffic directed to it's own IP addresses on _any_ interface. This can be adjusted with a sysctl, but that sysctl should not be configured by software as a security mechanism, as it may break a fair few things.  The proper(Er) solution is to add an interface-level filter for traffic.

This means that innernet client should also make sure to drop traffic directed to it's internal IP that arrives on any interface other than the wg interface.

SolomonSklash:
My understanding of Wireguard is that it provides this, not innernet itself:

"At the heart of WireGuard is a concept called Cryptokey Routing, which works by associating public keys with a list of tunnel IP addresses that are allowed inside the tunnel. Each network interface has a private key and a list of peers. Each peer has a public key. Public keys are short and simple, and are used by peers to authenticate each other. They can be passed around for use in configuration files by any out-of-band method, similar to how one might send their SSH public key to a friend for access to a shell server."

Innernet is a wrapper around associating IPs and interfaces and would not provide the authentication, Wireguard does.

DanielJoyce:
Unless the attacker can join the overlay it can't even see the destination ip. 

In order to join the wireguard overlay it needs a cert.

DanielJoyce:
Yes, traffic should be dropped that doesn't arrive from the proper interface.

mcginty:
[...]

PR #32, opened 2021-04-05
bind innernet-server specifically to WireGuard interface on Linux

mcginty:
This is one many upcoming changes to address IP spoofing issues.

See #26 for more details.

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -327,6 +327,12 @@ async fn serve(interface: &str, conf: &ServerConfig) -> Result<(), Error> {
     Ok(())
 }
 
+/// This function differs per OS, because different operating systems have
+/// opposing characteristics when binding to a specific IP address.
+/// On Linux, binding to a specific local IP address does *not* bind it to
+/// that IP's interface, allowing for spoofing attacks.
+///
+/// See https://github.com/tonarino/innernet/issues/26 for more details.
 #[cfg(target_os = "linux")]
 fn get_listener(addr: SocketAddr, interface: &str) -> Result<TcpListener, Error> {
     let listener = TcpListener::bind(&addr)?;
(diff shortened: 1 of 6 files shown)


server/src/main.rs
```diff
@@ -304,13 +306,43 @@ async fn serve(interface: &str, conf: &ServerConfig) -> Result<(), Error> {
...
     log::info!("innernet-server {} starting.", VERSION);
     let routes = routes(context.clone()).with(warp::log("warp")).boxed();
-    warp::serve(routes)
-        .run((config.address, config.listen_port))
-        .await;
+
+    let listener = get_listener((config.address, config.listen_port).into(), interface)?;
+
+    let warp_svc = warp::service(routes);
+    let make_svc = hyper::service::make_service_fn(move |socket: &AddrStream| {
```
skywhale, line 313: this is necessary because warp does not expose a way to use a user-specified `TcpListener`, I suppose?

mcginty: Yep, that's right - I also had to fork warp for now, pending https://github.com/seanmonstar/warp/issues/830.

server/src/main.rs
```diff
@@ -304,13 +306,43 @@ async fn serve(interface: &str, conf: &ServerConfig) -> Result<(), Error> {
...
+#[cfg(target_os = "linux")]
+fn get_listener(addr: SocketAddr, interface: &str) -> Result<TcpListener, Error> {
+    let listener = TcpListener::bind(&addr)?;
+    listener.set_nonblocking(true)?;
+    let sock = socket2::Socket::from(listener);
+    sock.bind_device(Some(interface.as_bytes()))?;
+    Ok(sock.into())
+}
+
+#[cfg(not(target_os = "linux"))]
```
skywhale, line 339: Might be obvious but a comment that `bind_device` is only supported for Linux, Android, and Fuchsia may be useful; in case we try to support other platforms in the future.

skywhale (approved):

server/src/main.rs
mcginty: Done!

PR #33, opened 2021-04-06
Remove some unsafe in the wgctrl-sys kernel backend

BlackHoleFox:
After reading the code a bit, I saw a bit of unsafe that caught my eye.

The previous device name parsing could potentially lead to UB since it doesn't look like Wireguard nor Netlink care about if the interface name is valid ASCII or UTF-8 and `str::from_utf8_unchecked` was being used. I made an `InterfaceName` wrapper struct to help validate interface names better and to ensure nothing weird happens passing the interface name around.

It also made converting the interface names cheaper since its just copies some bytes now. I also added some basic validation for interface name creation to prevent data that's going to be rejected anyway from making it across the FFI. As part of the newtype I moved the magic `16` constant to use the same `libc` constant that the Wireguard headers do for interfaces max length.

Also thanks for open sourcing this, its a very cool project and I look forward to using it in one of my setups :D

### DIFF
--- a/wgctrl-rs/src/device.rs
+++ b/wgctrl-rs/src/device.rs
@@ -2,7 +2,14 @@ use libc::c_char;
 
 use crate::{backends, key::Key};
 
-use std::{borrow::Cow, convert::{TryFrom, TryInto}, ffi::CStr, fmt, net::{IpAddr, SocketAddr}, time::SystemTime};
+use std::{
+    borrow::Cow,
+    convert::TryFrom,
+    ffi::CStr,
+    fmt,
+    net::{IpAddr, SocketAddr},
+    time::SystemTime,
+};
 
 /// Represents an IP address a peer is allowed to have, in CIDR notation.
 ///
@@ -141,6 +148,7 @@ impl TryFrom<&str> for InterfaceName {
 }
 
 impl InterfaceName {
+    #[cfg(target_os = "linux")]
     /// Creates a new [InterfaceName](Self).
     ///
     /// ## Safety
@@ -151,16 +159,20 @@ impl InterfaceName {
     }
 
     /// Returns a human-readable form of the device name.
-    fn as_str_lossy(&self) -> Cow<'_, str> {
+    ///
+    /// Only use this when the interface name was constructed from a Rust string.
+    pub fn as_str_lossy(&self) -> Cow<'_, str> {
         // SAFETY: These are C strings coming from wgctrl, so they are correctly NUL terminated.
         unsafe { CStr::from_ptr(self.0.as_ptr()) }.to_string_lossy()
     }
 
+    #[cfg(target_os = "linux")]
     /// Returns a pointer to the inner byte buffer for FFI calls.
     pub(crate) fn as_ptr(&self) -> *const c_char {
         self.0.as_ptr()
     }
 
+    #[cfg(target_os = "linux")]
     /// Consumes this interface name, returning its raw byte buffer.
     pub(crate) fn into_inner(self) -> RawInterfaceName {
         self.0
@@ -227,7 +239,7 @@ impl DeviceInfo {
     /// You can use [`get_by_name`](DeviceInfo::get_by_name) to retrieve more
     /// detailed information on each interface.
     #[cfg(target_os = "linux")]
-    pub fn enumerate() -> Result<Vec<String>, std::io::Error> {
+    pub fn enumerate() -> Result<Vec<InterfaceName>, std::io::Error> {
         if backends::kernel::exists() {
             backends::kernel::enumerate()
         } else {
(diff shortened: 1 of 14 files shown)


wgctrl-rs/src/device.rs
```diff
@@ -103,6 +108,121 @@ pub struct DeviceInfo {
...
 
+type RawInterfaceName = [c_char; libc::IFNAMSIZ];
+
+/// The name of a Wireguard interface device.
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub struct InterfaceName(RawInterfaceName);
+
+impl InterfaceName {
+    /// Creates a new [InterfaceName](Self).
+    pub fn new(name: &str) -> Result<Self, InvalidInterfaceName> {
```
mcginty, line 119: Thoughts on instead implementing this as `TryFrom` instead?

```rust
impl TryFrom<&str> for InterfaceName { ... }
```

That way all these methods that currently take a `interface: &str` can instead look like:

```rust
fn do_a_thing<I: TryInto<InterfaceName>>(interface: I)
```

which I think might make it easier to convert all cases where an interface is accepted as a `&str`.

mcginty (changes requested):
This makes me very happy to see, thanks so much for this PR. I really like how it looks, just a few notes.

Also, apologies but I didn't have GitHub Actions turned on for PRs. The macOS build doesn't compile with this change so we'll need to get that fixed before merging as well.

wgctrl-rs/src/device.rs
BlackHoleFox: I think that's reasonable. Ill convert it over and look at replacing more cases of `interface &str`

BlackHoleFox: @mcginty I moved it to a `TryFrom<&str>` implementation, but due to a lot of the methods returning `io::Result<()>`, you can't use it as a generic bound. Since when you perform the conversion in the function, it has no idea you're passing it the `&str` implementation so it is unable to convert `InvalidInterfaceType` into `io::Error`.

mcginty: Ah yeah you'd have to do something a bit more verbose like

```rust
pub fn get_by_name<T: TryInto<InterfaceName, Error = InvalidInterfaceName>>(
    name: T,
) -> Result<DeviceInfo, io::Error> {
    let name: InterfaceName = name.try_into()?;
    ...
}
```

Which is a bit much for every signature that takes that as an argument. I'd say let's keep it how you have it now and see if the code that uses `wgctrl-rs` becomes unwieldy before making the function signatures that messy :).

wgctrl-rs/src/device.rs
```diff
@@ -103,14 +109,137 @@ pub struct DeviceInfo {
     pub(crate) __cant_construct_me: (),
 }
 
+type RawInterfaceName = [c_char; libc::IFNAMSIZ];
+
+/// The name of a Wireguard interface device.
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub struct InterfaceName(RawInterfaceName);
+
+impl TryFrom<&str> for InterfaceName {
```
mcginty, line 118: I'm very sorry for suggesting `TryFrom<&str>` here, I think [`str::FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) is more flexible, since it can be easily used in `structopt`, for example. Thankfully it's basically the same signature. My bad.

wgctrl-rs/src/device.rs
```diff
@@ -103,14 +109,137 @@ pub struct DeviceInfo {
...
+    type Error = InvalidInterfaceName;
+
+    fn try_from(name: &str) -> Result<Self, InvalidInterfaceName> {
```
mcginty, line 121: Might be worth adding a rustdoc here referring to the page describing these name limitations (or even just linking to https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/utils.c#n827), if anyone was wondering where this logic came from.

wgctrl-rs/src/device.rs
```diff
@@ -103,14 +109,137 @@ pub struct DeviceInfo {
...
+        }
+
+        let mut buf = [c_char::default(); libc::IFNAMSIZ];
+        // Check for interior NULs and other invalid characters.
+        for (out, b) in buf.iter_mut().zip(name.as_bytes()[..(len - 1)].iter()) {
+            if *b == 0 {
+                return Err(InvalidInterfaceName::InteriorNul);
+            }
+
+            if *b == b'/' || *b == b' ' {
```
mcginty, line 139:
```suggestion
            if *b == b'/' || b.is_ascii_whitespace() {
```

It looks like iproute2, for example, checks not just for a space character but any whitespace character via [`isspace(3)`](https://linux.die.net/man/3/isspace). We probably want to follow suit.

BlackHoleFox: Hadn't considered that but its probably smart to follow the C as closely as possible.

wgctrl-rs/src/device.rs
BlackHoleFox: I added it in a doc comment on the `from_str` method as a reference.

wgctrl-rs/src/device.rs
BlackHoleFox: No problem, I moved it to use `FromStr` instead.

BlackHoleFox:
Finally got the tests passing too /o/

mcginty (approved):
Looks great, thanks so much @BlackHoleFox!

The next validation challenge after this is that innernet networks need to be both valid WireGuard interfaces as well as valid hostnames, since the name is used as a namespace for DNS resolution, so I'll try to tackle that today now that your change is merged :).

PR #35, opened 2021-04-06
Changeable hosts file settings

### LINKED ISSUE #6: Please don't edit /etc/hosts
Thank you for this wonderful project, this sounds exactly like what I wanted (something like Tailscale, but without a proprietary server part, in Rust, using the in-kernel WireGuard module! 🎉)

There's one thing stopping me from adopting Innernet immediately: from the glance at the client source code, it appears that Innernet edits `/etc/hosts` in order to enable `.wg` names to work.

In my book, that's very wrong. `/etc/hosts` is the configuration written and maintained by me, the system administrator, not by any software.

Off the top of my head, the two better ways to add your hostnames to the system are:
* make a glibc NSS module (like nss-mymachines)
* run a special DNS server on the WireGuard interface; tell the rest of the system (i.e. with `resolvconf` or systemd-resolved API) to use it for the `wg` domain.

mcginty:
Fixes #6, but could be further improved to have a persistent setting in a config file i.e. `/etc/innernet.conf` (which doesn't currently exist).

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -36,14 +36,14 @@ struct HostsOpt {
     #[structopt(long, default_value = "/etc/hosts")]
     hosts_path: PathBuf,
 
-    /// Don't modify /etc/hosts.
+    /// Don't write to any hosts files. This takes precedence over --hosts-path.
     #[structopt(long)]
     no_write_hosts: bool,
 }
 
 impl From<HostsOpt> for Option<PathBuf> {
     fn from(opt: HostsOpt) -> Self {
-        opt.no_write_hosts.then(|| opt.hosts_path)
+        (!opt.no_write_hosts).then(|| opt.hosts_path)
     }
 }
 
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -31,10 +31,32 @@ struct Opt {
 }
 
 #[derive(Debug, StructOpt)]
+struct HostsOpt {
+    /// The path to write hosts to.
+    #[structopt(long, default_value = "/etc/hosts")]
+    hosts_path: PathBuf,
+
+    /// Don't modify /etc/hosts.
+    #[structopt(long)]
+    no_write_hosts: bool,
+}
+
+impl From<HostsOpt> for Option<PathBuf> {
+    fn from(opt: HostsOpt) -> Self {
+        opt.no_write_hosts.then(|| opt.hosts_path)
+    }
+}
+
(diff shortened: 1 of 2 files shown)


jedahan:
If someone passes `--no-write-hosts --hosts-file /some/path/to/hosts` whats the result?

mcginty:
@jedahan `--no-write-hosts` takes precedence. That should be in the help text, thanks for the comment.

I could also just remove that flag and users would need to do `--hosts-file /dev/null`, but that feels a bit verbose to me, personally.

jedahan:
The inverse is another option - only write hosts if `--hosts-file` or something more explicit like `--write-hosts-to /path` exists. Then `--no-write-hosts` would be unnecessary.

The downside, of course, is losing the magic 'defaults just work' aspect, which seems a regression.

Either way, thanks for the thoughtful implementation and responses.

client/src/main.rs
```diff
@@ -30,11 +30,33 @@ struct Opt {
     command: Option<Command>,
 }
 
+#[derive(Debug, StructOpt)]
+struct HostsOpt {
+    /// The path to write hosts to.
+    #[structopt(long, default_value = "/etc/hosts")]
+    hosts_path: PathBuf,
+
+    /// Don't modify /etc/hosts.
```
skywhale, line 39: `Don't modify the hosts file. This takes precedence to the hosts_path option`

client/src/main.rs
```diff
@@ -30,11 +30,33 @@ struct Opt {
...
+    #[structopt(long)]
+    no_write_hosts: bool,
+}
+
+impl From<HostsOpt> for Option<PathBuf> {
+    fn from(opt: HostsOpt) -> Self {
+        opt.no_write_hosts.then(|| opt.hosts_path)
```
skywhale, line 46: Should this be `(!opt.no_write_hosts)`?

skywhale:
Thank you so much for the constructive feedback :) I think I'm leaning towards keeping the original behavior, until we have more data about which usage is more popular.

client/src/main.rs
mcginty: Thanks!

mcginty:
I'd also strongly prefer to keep the default hosts magic, since I find it to be one of the most useful aspects of innernet, personally.

I'll add a bit on the `install` finish message though noting that this is what's happening so people don't feel blindsighted.

Issue #50, opened 2021-04-13
Unable to compile on RPi4 (armhf, 32bit)


tlan:
I'm trying to compile innernet on RPi4 (Ubuntu, 20.04, 32bit ARMv7), but fails.  I've been trying with these two toolchains:

armv7-unknown-linux-gnueabihf
armv7-unknown-linux-musleabihf

(I really need the latter one, because I need this to be statically linked as I'm going to use it on another ARMv7 install in the end).

armv7-unknown-linux-gnueabihf fails on:
```
   Compiling wgctrl v1.1.0 (/home/ubuntu/src/innernet/wgctrl-rs)
error[E0308]: mismatched types
   --> wgctrl-rs/src/device.rs:142:20
    |
142 |             *out = *b as i8;
    |                    ^^^^^^^^ expected `u8`, found `i8`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0308`.
error: could not compile `wgctrl`
```

armv7-unknown-linux-musleabihf fails on:
```
   Compiling wgctrl-sys v1.1.0 (/home/ubuntu/src/innernet/wgctrl-sys)
error: failed to run custom build command for `wgctrl-sys v1.1.0 (/home/ubuntu/src/innernet/wgctrl-sys)`

Caused by:
  process didn't exit successfully: `/home/ubuntu/src/innernet/target/release/build/wgctrl-sys-050e5c811584b48d/build-script-build` (exit co
de: 101)
  --- stderr
  /usr/include/sys/types.h:144:10: fatal error: 'stddef.h' file not found
  /usr/include/sys/types.h:144:10: fatal error: 'stddef.h' file not found, err: true
  thread 'main' panicked at 'Unable to generate bindings: ()', wgctrl-sys/build.rs:15:44
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

bschwind:
Hmmm, probably this line:

https://github.com/tonarino/innernet/blob/a1818d9618521797f8816bd34b90992ed8d94c2c/wgctrl-rs/src/device.rs#L142

should be:

```rust
*out = *b as c_char;
```

What do you think, @mcginty ?

tlan:
Is there a way to see how it tries to find all the includes when compiling?  That way I can try to figure out if there's something that can fix one of the issues, at least.

mcginty:
Yes that's right! Sorry for the delay @tlan and thanks for opening the issue.

blesson3:
[...]

PR #54, opened 2021-04-18
use proper c char types

blesson3:
related: https://github.com/tonarino/innernet/issues/50

helps with building on systems like Raspberry Pi (ARM)


mcginty:
Thanks!

Issue #42, opened 2021-04-09
Support for Container/Kubernetes environments?


christianhuening:
Hi, we just found innernet and so far like it a lot. We're wondering whether you already have any plans or setups to run this in Kubernetes environments? We are thinking about using innernet to hide our development and production clusters within a private company network. 
We would go ahead and produce a dockerfile and a helm chart to try it, but just wanted to ask first, whether it has been done already somewhere.

mcginty:
Hey @christianhuening! Thanks I'm glad you like it so far! We don't use kubernetes internally so I haven't tested it, *but* it should work just fine in a Docker container if the host has the WireGuard kernel extension available, and if not you can instead run `wireguard-go` inside the container.

Something like https://github.com/cmulk/wireguard-docker (or https://github.com/masipcat/wireguard-go-docker for the second option) might be a good starting point for making your `Dockerfile` - in the next week or so I plan on adding better automated testing to this repo using either Docker or QEMU too, so there may be more news on that front in a bit.

Long story short, it should work as long as you're comfortable with adding the additional necessary `NET_ADMIN` capability for creating network interfaces on the host.

alesti:
Hi, we (Christian is one of my coworkers) appreciate your fast responses - it makes us really happy and also confident that we will use innernet in future.

I will try to explain a little bit more our docker (or better kubernetes) specific problems, we try to handle:

We are on the way to hide our public, but non workload endpoints of our clusters (api server, alertmanager, grafana, metrics, ...)

We know tailscale, but is very expensive and we cannot run our own yellow pages server - we work in a high regulated sector and try to host most stuff by ourselfs.
We checked [pritunl](https://docs.pritunl.com/docs), which uses a mongodb as configuration data store and has openvpn and wireguard as vpn protocols. It has a nifty webinterface to configure anything. It is

alesti:
Hi, we (@christianhuening  is one of my coworkers) appreciate your fast responses - it makes us really happy and also confident that we can use innernet in future.

I will try to explain a little bit more our docker (or better kubernetes) specific problems, we try to handle:

We are on the way to hide our non workload endpoints of our clusters (api server, monitoring, metrics, ...) from public.

We know tailscale, but it is very expensive and we cannot run our own control server - we work in a heavly regulated sector and try to host most of the at least security related stuff by ourselfs.

[...]

Issue #14, opened 2021-03-31
connect to network from docker container?


justinrush:
I'd like to use this to replace some openvpn site-to-site madness but two of the sites only have servers that are kubernetes nodes. I can see some cool ways to run this - like having an innernet server in each k8s cluster, having a sidecar that does some sort of on-start-get-priv-key thing from the local innernet server to join the network. However, it appears the way the client tries to run is not permitted inside a docker image.

From a windows host:
Using ubuntu:20.04, I get this error when trying to add a network:
```
root@f513c653fc32:~# inn install invite.toml
✔ Interface name · test
[*] bringing up the interface.

[ERROR] Operation not permitted (os error 1)
```

From a linux host:
```
root@074863c023a2:~# inn install /mydocker.toml
✔ Interface name · test
[*] bringing up the interface.

[ERROR] No such file or directory (os error 2)
```

thoughts on if this could be possible?

mcginty:
I'm wondering if this is related to https://github.com/tonarino/innernet/issues/5 - my understanding is that docker should work with WireGuard if the host's kernel has it.

Is it possible you just need to `modprobe wireguard`?

justinrush:
> I'm wondering if this is related to #5 - my understanding is that docker should work with WireGuard if the host's kernel has it.
> 
> Is it possible you just need to `modprobe wireguard`?

I saw that one - I am above 5.6. I can reprovision the VM and try to manually modprobe if we think thats helpful.

I assume this means running docker on windows just won't work? not an issue for me, it was just how i was testing something, all my actual targets are linux.

mcginty:
The more portable option is to use the wireguard-go userspace implementation instead of the kernel module, which works on Linux and should work in any docker container regardless of host modules.

kamiller8118:
I dont know what any of that is our how to find it,  i am completely new to everything

pilik88:
To tak jak ja, staram się zrozumieć cokolwiek...

[pilik88 - Chat @ Spike](https://spikenow.com/r/a/?ref=spike-organic-signature&_ts=yz2a3)	[yz2a3]

On April 3, 2021 at 8:13 GMT, Kris ***@***.***> wrote:

I dont know what any of that is our how to find it, i am completely new to everything

mcginty:
Check out the examples in #55 for details, this is now possible :).

PR #55, opened 2021-04-19
Docker-based integration tests

mcginty:
Related to https://github.com/tonarino/innernet/issues/14 and https://github.com/tonarino/innernet/issues/42.

This may also serve as an example of how to use innernet in a Docker environment, in lieu of an official image.


network-tests.sh
```diff
@@ -0,0 +1,29 @@
...
+tmp_dir=$(mktemp -d -t innernet-tests-XXXXXXXXXX)
+cleanup() {
+    rm -rf "$tmp_dir"
+    docker stop $(docker ps -q)
+    docker network remove innernet
+}
+trap cleanup EXIT
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    # load WireGuard's kernel module is loaded in the macOS Docker VM.
```
skywhale, line 13: nit: typo

network-tests.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
skywhale, line 1: It may be good to include `docker` in the script name.

PR #59, opened 2021-04-22
Fix cargo install instructions

dbr:
"git checkout" -> "git clone" and install both client and server for consistency with other install methods


mcginty:
Thanks!

PR #71, opened 2021-05-08
make adding routes optional

### LINKED ISSUE #70: allow for other routing methods
Currently innernet does static routing using the OS's built in tools in `shared/src/wg.rs`'s `add_route` function which is a good default functionality. However, if this functionality could be turned of somehow (not necessarily at runtime), that would enable users to choose other routing methods/protocols such as [babel](https://github.com/jech/babeld/).

Making innernet do that configuration would probably be a bit much, so by just switching this behaviour off (e.g. with a feature flag at compile time), people could configure their other routing methods themselves.

If this sounds like a reasonable idea, I can prepare a pull request on this.

Johann150:
Will resolve #70.

External routing can be enabled for the entire network when creating the server by using `innernet-server new --external-routing`. 

I did not add tests for this because those would have to be cross platform and I do not know how to do that. But ~~I did have to adjust one test and~~ all existing tests pass.

Something that I did not think through at the start is that for most dynamic routing protocols, users would probably need multiple wireguard interfaces - one per peer - so the routing daemon can operate on those. But that would be a much bigger change than this and out of scope. Maybe adding capabilities to make the network more into a mesh would be interesting in the future to reduce single points of failure - the server currently being one.

See some more comments below.


shared/src/interface_config.rs
```diff
@@ -50,6 +50,11 @@ pub struct ServerInfo {
 
     /// An internal endpoint in the WireGuard network that hosts the coordination API.
     pub internal_endpoint: SocketAddr,
+
+    /// Whether the routing should be done by innernet or is done by an
+    /// external tool like e.g. babeld.
+    /// Analog to what wg-quick does with "Table = off" if true.
+    pub external_routing: bool,
```
Johann150, line 57: It was not clear to me where the line is drawn between `ServerInfo` and `InterfaceInfo`. I added the `external_routing` to `ServerInfo` because that is what is already available in `add_peer` in `client/src/main.rs`.

shared/src/prompts.rs
```diff
@@ -289,6 +289,7 @@ pub fn save_peer_invitation(
     keypair: KeyPair,
     server_api_addr: &SocketAddr,
     config_location: &Option<String>,
+    external_routing: bool,
```
Johann150, line 292: Clippy now complains about too many arguments here. This might be solved by constructing the `InterfaceConfig` before calling the function.

Johann150 (commented):
These changes add a new failure condition when redeeming an invite, if the routing is not set up properly or has not registered the change quickly enough. This is not specially handled, the existing generic error message should be enough: https://github.com/tonarino/innernet/pull/71/files#diff-569c4866ec743e47bcc4258ec42720e55d00c382c8bc859b2916a4fe4f2ae030R247-R251

shared/src/prompts.rs
mcginty: That's ok, I can refactor that in another commit.

mcginty (changes requested):
Thanks for the PR!

I'd like to start with this feature making no modifications to any file formats, and instead only existing as an added CLI flag, which can be done kind of similarly to [`HostsOpt`](https://github.com/tonarino/innernet/blob/main/client/src/main.rs#L33), in that you can make that struct and add
```rust
struct RoutingOpt {
    /// Whether the routing should be done by innernet or is managed by an
    /// external tool like e.g. babeld.
    /// Analogous to what wg-quick does with "Table = off" if set.
    #[structopt(long)]
    no_routing: PathBuf,
}
...
enum Commands {
    Up {
        ...

        #[structopt(flatten)]
        routing: RoutingOpt
    },
    ...
}
```
to any subcommand that needs that flag, so that you don't have to copy-paste the rustdocs to a ton of places. In this case, since you may want either the server, client, or both to not do routing, this type could live in the `shared` crate.

Also, I have a preference towards calling the flag `--no-routing` instead of `--external-routing`, as the "external" implies that innernet will somehow talk to some external router, when in fact it's simply *not* doing something in this case.

Thanks again!

Johann150:
> I'd like to start with this feature making no modifications to any file formats, and instead only existing as an added CLI flag

That would mean I have to add that flag every time I run a command that `up`'s the wireguard interface, right? I think that would be inconvenient. (And by the way similar to the `--no-write-hosts` flag.) But I understand that what I am doing here is not your primary goal for this tool.

mcginty:
> > I'd like to start with this feature making no modifications to any file formats, and instead only existing as an added CLI flag
> 
> That would mean I have to add that flag every time I run a command that `up`'s the wireguard interface, right? I think that would be inconvenient. (And by the way similar to the `--no-write-hosts` flag.) But I understand that what I am doing here is not your primary goal for this tool.

That's correct, and I definitely agree that it's inconvenient. That said, I'd like to think a bit more about how machine-wide configurations are stored, since there's not currently a file for it on either the client or server side, and I want to hold back on throwing persistent options/preferences in the existing files for now.

For convenience, as with the `--no-write-hosts` option, for now I recommend editing the systemd units to add those arguments on, and if you're interested in working on the persistent preferences file (i.e. `/etc/innernet.conf`, `/etc/innernet-server.conf`), we can tackle that in a separate issue/PR.

PR #76, opened 2021-05-11
containers are annoying

adamgoose:
This properly spits out logs. Still failing, but hopefully helps debugging. Example build output: https://github.com/adamgoose/innernet/runs/2553015881?check_suite_focus=true


mcginty:
Hah, thanks! Good call 🙂.

Issue #19, opened 2021-03-31
Support Windows


jrop:
I saw your blogpost broadcasted on the Nebula Slack, and I must say, I am extremely interested in this project.  Also, I'm a Rust fanboy, so this project gets bonus points there as well 😛 .

Now for my question: Is Windows support planned?  I mostly use Linux/macOS, but there are instances where I want to connect with a Windows machine as well.

Eeems:
I've just finished installing this in WSL, so windows is _technically_ supported. Having it work on the host OS would be great though.

jrop:
@Eeems That is good to know. The use-case I'm thinking of, however, would be along the lines of giving this to a friend who may not be technically advanced and not even know about the existence of WSL.

DanielJoyce:
They'd need to support the wireguard Go implementation which is user space and what tailscale uses on non linux devices.

mcginty:
@Eeems that's great to hear! Thanks for reporting back on that here.

None of us at tonari are Windows users at the moment which is why it's unsupported now, but I'm interested in adding it. There's no timeframe on that, but I'd be happy to help anyone who's interested in submitting a PR.

Eeems:
In case anybody comes here and wants instructions on how to install in WSL: https://eeems.website/innernet-in-wsl

Johann150:
What other things would have to be done to run on Windows? I see there are already some things in the `wgctrl` crates for `not(target_os = "linux")`.

RSteendam:
I'd really like this feature as well. We use Wireguard in our company, but like the security and manageability this offers.

readall:
Is there a way that importing the configuration data to other wireguard client can work. I have on a windows workstation Wireguard client installed which uses wireguard-go. But when i try to import the data, the handshake keeps failing.

wwalker:
@readall Did you ever get this working?

PizzaProgram:
There are [many solutions](https://github.com/anderspitman/awesome-tunneling#overlay-networks-and-other-advanced-tools) out there, but less and less are supporting Windows 7 32 bit clients. :-(
(And by that, accelerating [this ...](https://duckduckgo.com/?t=ffab&q=africa+e-waste&iax=images&ia=images) )

If this could be an alternative to OpenVPN, it would be really great! 👍

ann4belle:
[...]

PR #79, opened 2021-05-11
adapt hostfile crate for Windows

Johann150:
Windows has some peculiarities for example it only allows one hostname per line and the file's location depends on an environment variable. Although in most cases just using `C:\Windows\` for `%WinDir%` would probably work, it is cleaner to use the documented path.

Note that editing the hosts file on Windows will require running with elevated privileges ("Run as Administrator") and in some cases also antivirus tools may block access to the file. (This is not currently handled specially - when I tried it, the error was the same for missing privileges, and for antivirus blocking access.)

This is a step towards #19.

### DIFF
--- a/hostsfile/src/lib.rs
+++ b/hostsfile/src/lib.rs
@@ -124,7 +124,7 @@ impl HostsBuilder {
                 // the location depends on the environment variable %WinDir%.
                 format!(
                     "{}\\System32\\Drivers\\Etc\\hosts",
-                    std::env::var("WinDir").expect("missing environment variable %WinDir%")
+                    std::env::var("WinDir")?
                 ),
             )
         } else {


hostsfile/src/lib.rs
```diff
@@ -106,11 +114,19 @@ impl HostsBuilder {
...
     pub fn write(&self) -> Result<()> {
         let hosts_file = if cfg!(unix) {
             PathBuf::from("/etc/hosts")
+        } else if cfg!(windows) {
+            PathBuf::from(
+                // according to https://support.microsoft.com/en-us/topic/how-to-reset-the-hosts-file-back-to-the-default-c2a43f9d-e176-c6f3-e4ef-3500277a6dae
+                // the location depends on the environment variable %WinDir%.
+                format!(
+                    "{}\\System32\\Drivers\\Etc\\hosts",
+                    std::env::var("WinDir").expect("missing environment variable %WinDir%")
```
mcginty, line 127: Let's return an `std::io::ErrorKind::NotFound` type error here instead of panicking.

mcginty (changes requested):
Looks great besides one small request, thanks!

hostsfile/src/lib.rs
Johann150: Would returning the `std::env::VarError` also be okay? That could be acomplished by just using the question mark operator and simpler than instantiating a different error.

mcginty: Ah, I forgot that this crate returns a generic `Box<dyn std::error::Error>`. Yes go ahead and use a `std::env::VarError`, and we can save the error refactoring for a later time 🙂.

Johann150: 👍 has been implemented that way

mcginty:
Thanks!

PR #84, opened 2021-05-18
Add Shell Completions

kbknapp:
These commits add a subcommand takes a shell as an argument and generates shell
completion scripts for that shell to stdout.

i.e.

```
$ innernet completions bash
  OR
$ innernet-server completions zsh
```

The supported shells:

- Bash
- ZSH
- Fish
- PowerShell
- Elvish

This is further used to generate some completion scripts for shells that
have standard locations to look for completion scripts by default (Bash and Fish)
at DEB package time.

--- 

Even without putting these into  packaging, a user could enable completions on the fly:

```
$ source <(innernet completions bash)
```

ZSH requires the completions be in a specific place under the user's
home dir, so the following would allow ZSH users access to
completions (assuming `~/.zfunc` is in their `$FPATH` or substituting
for an appropriate dir if not):

```
$ innernet completions zsh >
~/.zfunc/_innernet && exec zsh
```

## TODO

- [ ] ACTION: Someone more familiar with the packaging should ensure there isn't anything incorrect or missing, or point me in the right direction
- [x] DISCUSS: The packaging portion could be pulled out entirely if so desired.
- [x] DISCUSS: Another option is to check the scripts into the repo source control, and use them as a starting point that is further customized or tweaked/perfected.

Edit: adds another TODO bullet for discussing checking the completion scripts into source control instead of/in addition to packaging, or using a subcommand


mcginty (changes requested):
A fellow skydiver :)

This is great! Thanks so much for the PR. Small comments:

## Generated completion files
I'd much prefer to either:
1) Intentionally store the generated files in a location that's version controlled like the generated manpages in `doc/`, in which case we'd probably want to generate these in the `release.sh` script.
2) Generate the completions into `./target`  so that they can be safely considered a build artifact ignored by git.

I'd lean #1 since it's kind of nice to see how the completions have changed over time and it removed specific executable code needing to be added to packaging stuff, which will be a pain to maintain long-term.

## Aliases

How might this work with the `inn` alias? Is it as easy as renaming the completion file?

kbknapp:
> A fellow skydiver :)

Yep! Most people think my profile picture is a motorcycle helmet.

> [ Generated Completion files ]

I agree that option 1 is better. It allows you to tweak the scripts if needed. When the completion feature of clap was originally developed it was meant to do exactly this, and function as a starting point. However, it turned out to be good enough that people just use them as is without much real need for tweaking.

I'd still include the completions subcommand, which is nice in that it allows people who install through other means access to completion scripts in a pinch. If worried about polluting the CLI, we can also always just make the `completions` subcommand hidden so that it doesn't appear in the `--help` message.

> [ Aliases ]

This is depending on the shell. Both ZSH and Fish will transparently handle and give you the correct completions for aliased commands, even when using the alias. i.e. `$ inn --<tab>` will display the correct options. Bash however, does not. [There are hacky workarounds](https://unix.stackexchange.com/questions/4219/how-do-i-get-bash-completion-for-command-aliases), but this is at the user level, not the application level.  Unfortunately, renaming the completion script doesn't help with Bash, you have to also add a line to bottom of the completion script.

For example, if you `alias inn=innernet` you could run:

```
$ echo 'complete -F _innernet -o bashdefault -o default inn' >> /etc/bash_completion.d/innernet.sh
```

(replacing the `/etc/bash_completion.d/innernet.sh` with wherever you've saved/sourced your completion script).

There is no reason clap couldn't add that line for you when you've defined aliases on your top level command, but currently it does not as no one has asked for it.

i.e. 

```rust
#[derive(Debug, StructOpt)]
#[structopt(name = "innernet", about, alias = "inn")]
struct Opts { /* .. */ }
```

Would at least generate the appropriate line at the end of the Bash script. However, only for the `inn` alias, if someone used something else you'd be back in the same boat.

kbknapp:
One thing I forgot to add about when checking in the completion scripts themselves to the repository, one would need to keep them in sync with any CLI changes. If no tweaks are made to the checked in scripts, this syncing is super simple as a CI check could be performed to ensure they haven't changed and alert the committer if they have.

mcginty:
Option 1 sounds good! I'm not yet too concerned about the subcommand overload, I think we'll eventually want to re-organize the StructOpt layout or manually write our own usage text at some point as the commands build up, but that's for another day.

As for keeping the completions synced with code changes, I've been adding those types of tasks to the `release.sh` script, which is what is also responsible for re-generating the manpages at every release (but not every commit).

kbknapp:
Rebased on the latest master and adds the generate scripts to the repo.

The only two TBD item I'm still tracking is whether or not to include the Bash and Fish scripts in the DEB/RPM packaging, and I (or someone) still needs to ensure the DEB/RPMs grab/place those scripts correctly.

server/Cargo.toml
```diff
@@ -50,6 +50,8 @@ assets = [
   ["target/release/innernet-server", "usr/bin/", "755"],
   ["innernet-server@.service", "usr/lib/systemd/system/", "644"],
   ["../doc/innernet-server.8.gz", "usr/share/man/man8/", "644"],
+  ["server.bash", "etc/bash_completion.d/innernet", "644"],
+  ["server.fish", "usr/local/share/fish/vendor_completions.d/innernet-server.fish", "644"],
```
mcginty, line 54: Similarly, do we want to copy all the various completions in OS packages, or do we only want to choose a popular subset (eg. bash/fish/zsh)?

I think it makes sense that the completions that live in `doc/` and the completions installed in OS packages should be matching.

mcginty (changes requested):
Some small final comments, but looks great - I love how clap makes this code change so small by doing the magic for you.

rpm/Dockerfile
```diff
@@ -15,6 +15,10 @@ RUN rm -rf target
 RUN source $HOME/.cargo/env && \
 	cargo install cargo-rpm && \
 	cargo build --release --verbose&& \
+    ./target/release/innernet completions bash > client.bash && \
+    ./target/release/innernet completions zsh > client.zsh && \
+    ./target/release/innernet-server completions bash > server.bash && \
+    ./target/release/innernet-server completions zsh > server.zsh && \
```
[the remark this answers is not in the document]
kbknapp: You're correct, these are no longer needed with the ones in `doc/` I had forgot to look in the Dockerfile.

server/Cargo.toml
kbknapp: I only did Bash/Fish since they have standard locations where the OS looks for global completions. The other shells don't have a directory like that. So we could place the scripts for the shells without a global directory somewhere like `/usr/share/innernet/completions/` at which point it's up to the user to do something with those scripts if they wish. It'd just need to be documented somewhere that they're available. I just didn't want to make that assumption without approval first.

mcginty: It looks like zsh can be dropped in `/usr/share/zsh/site-functions/_innernet` (or `_innernet-server`). I think sticking to the holy trinity of zsh, fish, and bash in packages makes sense to me! Also in that referenced line 53, should that be `innernet-server` and not `innernet`?

mcginty: (i assumed this just looking at how `ripgrep` does its packages in arch: https://github.com/archlinux/svntogit-community/blob/packages/ripgrep/trunk/PKGBUILD#L36)

kbknapp: Sounds good to me, I'll make that change!

mcginty:
Looks great, thanks so much!

PR #88, opened 2021-05-20
adds ability to delete cidrs

### LINKED ISSUE #23: Allow deleting CIDR that has no peers
If you add a CIDR and realize that you made a mistake it would be good to allow removing the entry without having to edit the sqlite database.

kbknapp:
This commit adds a `delete-cidr` to both the client and server. It walks
through the prompts just like adding a CIDR. Only eligible CIDRs are shown to the user. 
Eligibility requirements are:

- CIDR has no child CIDRs
- CIDR has no assigned peers

Closes #23

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -4,9 +4,9 @@ use hostsfile::HostsBuilder;
 use indoc::printdoc;
 use shared::{
     interface_config::InterfaceConfig, prompts, AddAssociationOpts, AddCidrOpts, AddPeerOpts,
-    Association, AssociationContents, Cidr, CidrTree, EndpointContents, InstallOpts, Interface,
-    IoErrorContext, NetworkOpt, Peer, RedeemContents, State, CLIENT_CONFIG_DIR,
-    REDEEM_TRANSITION_WAIT,
+    Association, AssociationContents, Cidr, CidrTree, DeleteCidrOpts, EndpointContents,
+    InstallOpts, Interface, IoErrorContext, NetworkOpt, Peer, RedeemContents, State,
+    CLIENT_CONFIG_DIR, REDEEM_TRANSITION_WAIT,
 };
 use std::{
     fmt,
@@ -143,6 +143,14 @@ enum Command {
 
         #[structopt(flatten)]
         opts: AddCidrOpts,
+    },
+
+    /// Delete a CIDR.
+    DeleteCidr {
+        interface: Interface,
+
+        #[structopt(flatten)]
+        opts: DeleteCidrOpts,
     },
 
     /// Disable an enabled peer.
@@ -563,6 +571,23 @@ fn add_cidr(interface: &InterfaceName, opts: AddCidrOpts) -> Result<(), Error> {
     Ok(())
 }
 
+fn delete_cidr(interface: &InterfaceName, opts: DeleteCidrOpts) -> Result<(), Error> {
+    let InterfaceConfig { server, .. } = InterfaceConfig::from_interface(interface)?;
+    println!("Fetching eligible CIDRs");
+    let api = Api::new(&server);
+    let cidrs: Vec<Cidr> = api.http("GET", "/admin/cidrs")?;
+    let peers: Vec<Peer> = api.http("GET", "/admin/peers")?;
+
+    let cidr_id = prompts::delete_cidr(&cidrs, &peers, &opts)?;
+
+    println!("Deleting CIDR...");
+    let _ = api.http("DELETE", &*format!("/admin/cidrs/{}", cidr_id))?;
+
+    println!("CIDR deleted.");
+
+    Ok(())
+}
+
 fn add_peer(interface: &InterfaceName, opts: AddPeerOpts) -> Result<(), Error> {
     let InterfaceConfig { server, .. } = InterfaceConfig::from_interface(interface)?;
     let api = Api::new(&server);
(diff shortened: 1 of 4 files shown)


client/src/main.rs
```diff
@@ -563,6 +571,36 @@ fn add_cidr(interface: &InterfaceName, opts: AddCidrOpts) -> Result<(), Error> {
...
+    let peers: Vec<Peer> = api.http("GET", "/admin/peers")?;
+
+    let cidr_id = prompts::delete_cidr(&cidrs, &opts)?;
+
+    if cidrs.iter().any(|cidr| {
+        if let Some(parent_id) = cidr.contents.parent {
+            cidr_id == parent_id
+        } else {
+            false
+        }
```
mcginty, line 588: seems like a good time for `matches!`
```suggestion
        matches!(cidr.contents.parent, Some(parent_id) if parent_id == cidr_id)
```

shared/src/prompts.rs
```diff
@@ -58,6 +58,37 @@ pub fn add_cidr(cidrs: &[Cidr], request: &AddCidrOpts) -> Result<Option<CidrCont
     )
 }
 
+/// Bring up a prompt to delete a CIDR. Returns the peer request.
+pub fn delete_cidr(cidrs: &[Cidr], request: &DeleteCidrOpts) -> Result<i64, Error> {
+    let eligible_cidrs: Vec<_> = cidrs
+        .iter()
+        .filter(|cidr| cidr.name != "innernet-server")
+        .collect();
```
mcginty, line 66: In this case, for the sake of user-friendliness, we might want to filter the `eligible_cidrs` more here to *only* show CIDRs that the client thinks can reasonably be deleted (with the server having the final say, of course).

```suggestion
/// Bring up a prompt to delete a CIDR. Returns the peer request.
pub fn delete_cidr(cidrs: &[Cidr], peers: &[Peer], request: &DeleteCidrOpts) -> Result<i64, Error> {
    let eligible_cidrs: Vec<_> = cidrs
        .iter()
        .filter(|cidr| {
            !cidrs.iter().any(|c| c.parent == cidr.id) && peers.iter().any(|p| p.cidr_id == cidr.id)
        })
        .collect();
```

^ this suggested code probably doesn't compile - just there to show roughly what I'm thinking

server/src/main.rs
```diff
@@ -317,6 +327,41 @@ fn add_cidr(
...
+
+    let conn = open_database_connection(interface, conf)?;
+    let cidrs = DatabaseCidr::list(&conn)?;
+    let peers = DatabasePeer::list(&conn)?;
+
+    // let InterfaceConfig { server, .. } = InterfaceConfig::from_interface(interface)?;
+    // println!("Fetching CIDRs");
+    // let api = Api::new(&server);
+    // let cidrs: Vec<Cidr> = api.http("GET", "/admin/cidrs")?;
+    // let peers: Vec<Peer> = api.http("GET", "/admin/peers")?;
```
mcginty, line 340: can delete these :)

mcginty (changes requested):
Rad! Thanks so much for taking this on.

client/src/main.rs
kbknapp: Looks great, I wasn't aware of this macro!

shared/src/prompts.rs
kbknapp: I considered this as the first route, but I ran into a hiccup (although I don't remember what stopped me and caused me to just emit an error). I'll check out this code and make sure it compiles to integrate it, cause I agree not showing the ineligible cidrs at all is a better experience.

server/src/main.rs
kbknapp: Whooops :)

kbknapp:
Should be good to go now. Only eligible CIDRs are presented to the user.

kbknapp:
Added some fixes from testing locally. Squashed the commits down. Only shows eligible CIDRs now.

shared/src/prompts.rs
```diff
@@ -58,6 +58,43 @@ pub fn add_cidr(cidrs: &[Cidr], request: &AddCidrOpts) -> Result<Option<CidrCont
...
+            !cidrs.iter().any(
+                |cidr2| matches!(cidr2.contents.parent, Some(parent_id) if parent_id == cidr.id)
+            )
+        })
+        .collect();
+    let cidr = if let Some(ref name) = request.name {
+        cidrs
+            .iter()
+            .find(|cidr| &cidr.name == name)
+            .ok_or_else(|| format!("CIDR {} does not exist", name))?
```
mcginty, line 77:
```suggestion
            .ok_or_else(|| format!("CIDR {} doesn't exist or isn't eligible for deletion", name))?
```

shared/src/prompts.rs
```diff
@@ -58,6 +58,43 @@ pub fn add_cidr(cidrs: &[Cidr], request: &AddCidrOpts) -> Result<Option<CidrCont
     )
 }
 
+/// Bring up a prompt to delete a CIDR. Returns the peer request.
+pub fn delete_cidr(cidrs: &[Cidr], peers: &[Peer], request: &DeleteCidrOpts) -> Result<i64, Error> {
+    let eligible_cidrs: Vec<_> = cidrs
+        .iter()
+        .filter(|cidr| {
+            cidr.name != "innernet-server" &&
```
mcginty, line 66: I think we can remove this check for "innernet-server" here since it will never be eligible for deletion.

```suggestion
```

mcginty (changes requested):
Looks excellent - some tiny final requests then let's merge it!

shared/src/prompts.rs
kbknapp: Great point!

mcginty:
Thanks, merged!

Issue #41, opened 2021-04-09
Allow running on Linux without root


mcginty:
Linux users should be able to create an `innernet` user that has permissions to `/etc/innernet` and `/var/lib/innernet`, and the binaries should be runnable after giving it `CAP_NET_ADMIN` permission.

As far as I know, no such equivalent exists on macOS.

mcginty:
Yeah, I really wish Linux had a `CAP_NET_TUN` that was more constricted, but no such capability exists. If anyone was interested in creating SELinux profiles for innernet, though, I think that would be a great addition to the official documentation/packages.

mk-fg:
I've made this one to run it with AppArmor here:
```
#include <tunables/global>

profile innd /usr/bin/innernet-server {
  /usr/bin/innernet-server rm,
  /etc/innernet-server/ rw,
  /etc/innernet-server/* rw,
  /var/lib/innernet-server/ rw,
  /var/lib/innernet-server/* rwlk,

  /usr/lib/** rm,
  /etc/ld.so.* r,
  /dev/urandom r,
  /dev/null rw,
  owner @{PROC}/@{pid}/{maps,cgroup} r,

  network inet,
  capability net_admin,
  signal (receive) peer=unconfined,
}
```
SELinux one can probably have similar set of rules, just in its own syntax.

mk-fg:
> If anyone was interested in creating SELinux profiles for innernet, though, I think that would be a great addition to the official documentation/packages.

I've made this AppArmor profile to run innernet-server locally:
```
abi <abi/3.0>,
#include <tunables/global>

profile innd /usr/bin/innernet-server {
  /usr/bin/innernet-server rm,
  /etc/innernet-server/ rw,
  /etc/innernet-server/* rw,
  /var/lib/innernet-server/ rw,
  /var/lib/innernet-server/* rwlk,

  # Runs within same profile to work with NoNewPrivileges=yes
  /usr/bin/ip ix,
  /usr/bin/ip rm,

  /usr/lib/** rm,
  /etc/ld.so.* r,
  /dev/urandom r,
  /dev/null rw,
  owner @{PROC}/@{pid}/{maps,cgroup} r,

  network inet,
  network inet6,
  network netlink raw,
  network unix, # systemd journal stdout/stderr

  capability net_admin,
  signal (receive) peer=unconfined,
}
```

It does not include rules for running userspace wireguard, as I don't have it here.
[...]

PR #94, opened 2021-05-27
client: support running as non-root

mcginty:
This is a work-in-progress change to have the client support running as non-root, including better messaging around permission/access issues.

# Example configurations

## Linux
```sh
sudo cap_net_admin+eip $(which innernet)
sudo chown -R $USER /etc/innernet
sudo chown -R $USER /var/lib/innernet
```

## macOS
There are no "capabilities" in macOS, so subcommands like `up` and `install` will still require root.
```sh
sudo chown -R $USER /etc/innernet
sudo chown -R $USER /var/lib/innernet
sudo chown -R $USER /var/run/wireguard
```

### DIFF
--- a/hostsfile/src/lib.rs
+++ b/hostsfile/src/lib.rs
@@ -2,7 +2,7 @@ use std::{
     collections::HashMap,
     fmt,
     fs::{self, File, OpenOptions},
-    io::{BufRead, BufReader, Write},
+    io::{self, BufRead, BufReader, ErrorKind, Write},
     net::IpAddr,
     path::{Path, PathBuf},
     result,
@@ -115,7 +115,7 @@ impl HostsBuilder {
 
     /// Inserts a new section to the system's default hosts file.  If there is a section with the
     /// same tag name already, it will be replaced with the new list instead.
-    pub fn write(&self) -> Result<()> {
+    pub fn write(&self) -> io::Result<()> {
         let hosts_file = if cfg!(unix) {
             PathBuf::from("/etc/hosts")
         } else if cfg!(windows) {
@@ -124,18 +124,21 @@ impl HostsBuilder {
                 // the location depends on the environment variable %WinDir%.
                 format!(
                     "{}\\System32\\Drivers\\Etc\\hosts",
-                    std::env::var("WinDir")?
+                    std::env::var("WinDir").map_err(|_| io::Error::new(
+                        ErrorKind::Other,
+                        "WinDir environment variable missing".to_owned()
+                    ))?
                 ),
             )
         } else {
-            return Err(Box::new(Error("unsupported operating system.".to_owned())));
+            return Err(io::Error::new(
+                ErrorKind::Other,
+                "unsupported operating system.".to_owned(),
+            ));
         };
 
         if !hosts_file.exists() {
-            return Err(Box::new(Error(format!(
-                "hosts file {:?} missing",
-                &hosts_file
-            ))));
+            return Err(ErrorKind::NotFound.into());
         }
 
         self.write_to(&hosts_file)
@@ -146,7 +149,7 @@ impl HostsBuilder {
     ///
     /// On Windows, the format of one hostname per line will be used, all other systems will use
     /// the same format as Unix and Unix-like systems (i.e. allow multiple hostnames per line).
-    pub fn write_to<P: AsRef<Path>>(&self, hosts_path: P) -> Result<()> {
+    pub fn write_to<P: AsRef<Path>>(&self, hosts_path: P) -> io::Result<()> {
         let hosts_path = hosts_path.as_ref();
         let begin_marker = format!("# DO NOT EDIT {} BEGIN", &self.tag);
         let end_marker = format!("# DO NOT EDIT {} END", &self.tag);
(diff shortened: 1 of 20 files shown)


strohel:
I've tested this quickly. It complains nicely when the perms are not set:
```
strohel@mat480s ~/work/innernet/client $ innernet 
ERROR: innernet can't access its config/data folders.

You either need to run innernet as root, or give the user running innernet permissions
to access /etc/innernet and /var/lib/innernet.

[E] /var/lib/innernet - Permission denied (os error 13)
```
```
strohel@mat480s ~/work/innernet/client $ innernet 
ERROR: innernet can't access the device info.

You either need to run innernet as root, or give innernet CAP_NET_ADMIN capabilities:

    sudo setcap cap_net_admin+eip /home/strohel/.cargo/bin/innernet

[E] tonari - Operation not permitted (os error 1)
```
```
strohel@mat480s ~/work/innernet/client $ innernet up tonari

[E] /etc/innernet - Permission denied (os error 13)
```

`innernet show` started working afterwards, `down` worked too. `up` didn't:
```
strohel@mat480s ~/work/innernet/client $ innernet up --no-write-hosts tonari
[*] bringing up the interface.

[E] failed to run ip address replace 10.42.64.7/16 dev tonari command: RTNETLINK answers: Operation not permitted
```

Also, if running 2 concurrent innernet instances/devices by different users is fine, perhaps /etc/innernet, /var/lib/innernet can be replaced by XDG_CONFIG_HOME (`$HOME/.config`), XDG_(DATA or STATE)_HOME (`$HOME/.local/share` or `$HOME/.local/state`)?

kbknapp:
Does `CAP_NET_ADMIN` follow `execve` calls? Since `innernet` is calling out to other executables, it's possible those capabilities aren't following through the exec call since iproute2 isn't owned by `$USER`. I'm speculating here, and could be totally wrong.

danshick:
As a nitpick, I'm not sure I'd make anything in `/etc` owned by a user. Innernet should have no problem reading its config as a user initiated process if the permissions on the config are set to allow o+r. Either that, or read innernet's config from somewhere like `$XDG_CONFIG_HOME` and treat `/etc` as a fallback.

Similarly, I don't know what innernet uses `/var/lib` for, but I'd recommend making an `innernet` user and group, having that user and group own the directory in var, and suggesting that other users join the `innernet` group. Giving that whole directory g+rwx should be fine. Either that, or move innernet's data directory to somewhere like `$XDG_DATA_HOME`. https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

mcginty:
@kbknapp that sounds like it too - a good excuse for us to finally switch to using netlink directly via something like `netlink_sys` :)

@danschick - Thanks, and that makes a lot of sense! In the same model as WireGuard, the configs should be o+r since they contain the private key of the interface. I agree about a user not owning anything in `/etc`, but it _does_ seems like pretty common practice to have a folder like `/etc/innernet` owned by a non-root group like `innernet`, and I'd lean towards that approach for now over `$XDG_*` directories if only because innernet is still making system-wide changes and the configs should reflect that (i.e. there can't be two users with their own `foo` innernet interface, they'd conflict).

danshick:
> In the same model as WireGuard, the configs should be o+r since they contain the private key of the interface.

Ah, that makes sense. In that case, I'd suggest a file owner of root and a group of innernet with something like 770 (or 660 for private key material) as the file permissions for anything in /etc/innernet.

mcginty:
I decided to put the netlink-related changes in this PR as well, since it's related enough that it's worth testing as part of this larger non-root work.

Just replaced all execve calls to the `ip` tool with `netlink-sys` (FFI wrapper around `libnl`) calls and its associated helper crates - eventually we can do this for `wgctrl-rs` as well, and remove the `wgctrl-sys` crate all together.

Next up is fixing up the ownership issues, and then we should be close to good on this.

hostsfile/src/lib.rs
```diff
@@ -179,24 +174,19 @@ impl HostsBuilder {
...
-            .parent()
-            .expect("hosts file must be an absolute file path");
-        let temp_dir = tempfile::Builder::new().tempdir_in(hosts_dir)?;
-        let temp_path = temp_dir.path().join("hosts");
-
-        // Copy the existing hosts file to preserve permissions.
-        fs::copy(&hosts_path, &temp_path)?;
-
-        let mut file = File::create(&temp_path)?;
+        let mut file = OpenOptions::new()
```
skywhale, line 184: Do you want to write all the contents into a `String` first and `write!` that to the file? It should be atomic in most operating systems according to:
https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.append

Hosts file shouldn't be edit frequently, and race conditions are not realistic, but since this is a specific crate for editing the file, I think it makes sense to give a bit of consideration to concurrent writes. Another option would be to use a lock file.

mcginty: That makes sense, thanks for looking!

mk-fg:
@kbknapp
> Does CAP_NET_ADMIN follow execve calls? Since innernet is calling out to other executables, it's possible those capabilities aren't following through the exec call since iproute2 isn't owned by $USER. I'm speculating here, and could be totally wrong.

It depends on which capability sets are used - if that bit is only set in `permitted` and `effective` sets, then it will be reset on exec. If it's also set in `inheritable` set, then binary file would need to have capability bits set in its xattr for that bit to be preserved in any way (and it might be inherited only as `permitted` depending on these).
If it is in `ambient` set, then it'll just always be inherited across exec, e.g. like unit file in https://github.com/tonarino/innernet/issues/41#issuecomment-837932991 does.

Using netlink directly should avoid needing ambient capability though, so any random/bogus exec will properly loose it.

PR #96, opened 2021-05-29
Exit with code 1 (error) if wizard failed.

amarao:
I started to automate (1.3.0) and found that my code is 'ok' when there are errors in stdout.
I found that message 
`creation failed: failed to create database (are you not running as root?).`
is not causing non-zero exit code.

This fix should handle all `init_wizard` errors.

Also, the error message now is printed to stderr instead of stdout.


kbknapp:
Tangentially related, it should probably also print this to stderr (`eprintln!`) instead of stdout (`println!`)

amarao:
Yes, you're right. Updated PR.

mcginty:
Thanks so much for your issues and PRs around automation, it's something I definitely want to improve. Sorry for the delayed response!

PR #127, opened 2021-08-05
Fix clippy warnings, add clippy to CI

strohel:
#### Tidy code a bit thanks to clippy

Clippy 1.54 newly detects some redundant constructs, that's nice.

sort_unstable() should yield exact same results as sort() for `Vec<&str>`
and could be faster, clippy says.

#### Add clippy to CI


mcginty (approved):
thanks!

Issue #126, opened 2021-08-05
Refactoring of client code into a library


bramvdbogaerde:
Hello there, I am looking to integrate a GUI with this project to make connections to an innernet network by less technical people possible.

However, it seems to me that the `client` binary crate contains a lot of code for maintaining Wireguard devices, as well as communication with the remote API. In the interest of creating more interaction surfaces (e.g., GUIs, Android applications, ...), I think it would be beneficial to refactor some of this code into its own library, separated from the command line interface code. 

I would be interested to help with such refactoring, but I haven't had time to look into the code much, so I could be wrong and perhaps my suggestion does not really fit in the design philosophy of the project.

mcginty:
Hey @bramvdbogaerde, very sorry for the late reply! I would *love* to help you work on GUIs for innernet, including refactoring the client code to provide an easy API for different frontends.

bschwind:
I believe this has (mostly?) been done by #382, so I'll opportunistically close this.

mbernat:
Well, the work has been started there but it only contains a few APIs so far, most of the CLI functionality is still missing.

bschwind:
Oh whoops, we can reopen if you'd like, sorry.

mbernat:
No worries, it's an ancient issue anyway. Just a note for future visitors coming this way :)

alerque:
It would be useful to have this issue *open* until the library extraction process is relatively feature complete. I have it on a watch list because it should enable another project I want to try, but I can't track the status of a closed issue. The status change to 'closed' should reflect the actual status of the issue.

PR #132, opened 2021-08-18
Refactoring of client code into a library 

bramvdbogaerde:
This pull request tracks the progress on refactoring the Innernet client side (see issue #126).

It proposes the following changes:

* [x] Rename the `client` crate to the `cli` create to make the distinction between multiple frontend clearer
* [x] Create a `client` library crate which will host the code shared between frontends
* [ ]  Move the API calls into the `client` library
* [ ]  Move common code (such as the loop that periodically updates peers, adding peers, ...) into the `client` library
* [ ]  Create a `Store` crate which can be used to keep track of Innernet's local state in a platform and frontend agnostic way, the sole implementation of this trait will initially be the already implemented file backed store 

These are the initial proposed set of changes, it should make it easier to extend Innernet with more interaction surfaces that can be decoupled from the main Innernet codebase.

Comments and feedback are greatly appreciated.


mcginty:
Thanks for kicking this off! I think this is the right way forward, including breaking up the client into a couple of logically separated crates (and moving some of the current workspace crates out into their own repositories and putting them on crates.io).

mcginty:
I'm going to go ahead and close this until there's an active effort. Thanks for getting the conversation started @bramvdbogaerde!

bramvdbogaerde:
Indeed, I didn't have the time to progress this much further. I might reopen it in the future.

Issue #115, opened 2021-07-17
NAT holepunching code?


heyheyhello:
Hi @mcginty 

The innernet blog post mentioned NAT holepunching:

> We had some simple goals:
>  - Conveniences a typical WireGuard user wants: peer names, auto-updating peer lists, groups based on IP blocks, and automatic NAT holepunching. 

I had just read Tailscale's great blog post about NAT holepunching mechanisms and learned of the many approaches to the problem. I was curious how innernet did it but only found this in the codebase:

https://github.com/tonarino/innernet/blob/c01c2be4bb5b7782ec8a3ae858cd2e647011465b/server/src/api/mod.rs#L8-L14

Is there more implementation than this? I tried searching for "endpoint" in the codebase to find relevant NAT table code or port scanning code - I found  `persistent_keepalive_interval` and `spawn_endpoint_refresher` but it seems to be only part of innernet-server and not the clients/peers...

I looked up if Wireguard does holepunching only found [example code of how it _could_ be done](https://git.zx2c4.com/wireguard-tools/tree/contrib/nat-hole-punching/README); with a note to never use the code haha.

#109 wants to implement more NAT code but does innernet do any NAT holepunching on its own right now?

Maybe the Tailscale blog hyped it up as a more difficult thing than it is - in Nebula's code it looks to be a simple for-loop that send 1 byte to every peer and then sleeps: https://github.com/slackhq/nebula/blob/785914071104c73515736aafd2b9d91501108b23/hostmap.go#L369

Thanks for clarification!

mcginty:
hey @heyheyhello! Thanks for opening the issue, this is a great time to write a better explanation of where we're at and what the current limitations are.

Here's what innernet currently does:
1. Peers connect to the server via WireGuard (the server must have its WireGuard port open).
2. The server, via that `inject_endpoints` function you referenced, gossips the endpoint *it* sees for each peer to all their associated peers.
3. Peers then set their associated peers' endpoints to what the server told them.
4. Since `persistent_keepalive` is set for each peer to send a heartbeat packet every 25 seconds, peers behind NATs will be sending UDP holepunching packets (via the keepalive functionality).

We're able to avoid the need to use raw packets, since we are in fact talking to the server over WireGuard, so we don't need to use BPF to spoof the source port like they do in the holepunching example from the WireGuard repository.

[...]

Issue #109, opened 2021-07-01
NAT traversal [tracking issue]


strohel:
This is a tracking issue for applying various techniques in innernet to traverse NATs in the path. CC @skywhale.

@mcginty has curated https://tailscale.com/blog/how-nat-traversal-works/ as the best resource on toolbox we have for NAT traversal (and I second).

Some aux pieces information to start:

Q: If I manually port-forward innernet port on my side, that could increase chances to connect to that other peer behind NAT?
A: Yep, setting up port forwarding for innernet should fix your issues. Call `innernet set-listen-port` to get a stable port configure.

Q: Just curious, how does innernet determine my own NAT-translated port? With the help of the server?
A: Yep! the server acts as STUN server, basically

mcginty:
Added more details in https://github.com/tonarino/innernet/issues/115#issuecomment-882433652 that help describe our current state of NATitude.

mcginty:
## Phase 1 (STUN)

#118: Add awareness of the likelihood of a working peer endpoint and don't overwrite the endpoint when the peer is already connected. This allows our current basic NAT traversal to work properly.

## Phase 2 (ICE)

1. Modify the state protocol to allow clients to report a list of self-discovered candidate endpoints, and for the server to pass those along to other peers.
2. Modify the client's `fetch()` to be able to test multiple candidate IPs as endpoints and settle on the one that appears to be correct.

## Phase 3 (TURN)

This is lower-priority since it adds complexity and can cause larger fluctuations in network performance if the logic isn't well-designed.

1. Add support for relay nodes, which is probably similar to the site-to-site proposals seen before.

brandonkal:
Wiretrustee (another open source tailscale alternative) has this implemented and interestingly is missing but planning for the key management features you have here. It may be useful as a reference, and ultimately worth looking into merging efforts.

https://github.com/wiretrustee/wiretrustee

brandonkal:
## More reference implementations

https://docs.rs/p2p/0.6.0/p2p/

> The message to the rendezvous servers includes our public key so the message they send back is encrypted. This prevents some routers/firewalls from identifying it's a rendezvous attempt by looking at the message body and thus either mangling or discarding the packet. Such routers/firewalls seem to scan for socket addresses in the body and if it matches the ones in the router's pool they try to figure out it's a rendezvous/STUN attempt. With encrypted contents there is no chance of such detection, so we are safe there.

https://github.com/dswd/vpncloud
[...]

PR #134, opened 2021-08-25
NAT traversal: ICE-esque candidate selection

mcginty:
This change adds the ability for peers to report additional candidate endpoints for other peers to attempt connections with outside of the endpoint reported by the coordinating server.

While not a complete solution to the full spectrum of NAT traversal issues (TURN-esque proxying is still notably missing), it allows peers within the same NAT to connect to each other via their LAN addresses, which is a win nonetheless. In the future, more advanced candidate discovery could be used to punch through additional types of NAT cone types as well.

Related discussions #109 and #115


shared/src/netlink.rs
```diff
@@ -43,14 +46,31 @@ fn netlink_call(
...
 
-    let n_received = socket.recv(&mut buf[..], 0).unwrap();
-    let response = NetlinkMessage::<RtnlMessage>::deserialize(&buf[..n_received])
-        .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
-    log::trace!("netlink response: {:?}", response);
-    if let NetlinkPayload::Error(e) = response.payload {
-        return Err(e.to_io());
+    let mut responses = vec![];
+    loop {
+        let n_received = socket.recv(&mut buf[..], 0).unwrap();
```
strohel, line 51: do we want to keep the `.unwrap()`s on netlink socket, or shall we convert them to `?`? I guess IO errors don't really happen on the netlink socker, but still.

shared/src/netlink.rs
```diff
@@ -43,14 +46,31 @@ fn netlink_call(
...
+        let mut offset = 0;
+        loop {
+            let bytes = &buf[offset..];
+            let response = NetlinkMessage::<RtnlMessage>::deserialize(bytes)
+                .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
+            responses.push(response.clone());
+            log::trace!("netlink response: {:?}", response);
+            match response.payload {
+                // We've parsed all parts of the response and can leave the loop.
+                NetlinkPayload::Ack(_) | NetlinkPayload::Done => return Ok(responses),
```
strohel, line 61: Do we want the final `Ack`/`Done` be part of the returned `responses`? (order of the `.push()` vs. the `match {}`)

shared/src/types.rs
```diff
@@ -570,7 +567,10 @@ impl<'a> PeerDiff<'a> {
         }
 
         // We won't update the endpoint if there's already a stable connection.
-        if !Self::peer_recently_connected(&old_info) {
+        if old_info
+            .map(|info| !info.is_recently_connected())
+            .unwrap_or(true)
```
strohel, line 572: What is easier for your brain?
```suggestion
        if !old_info
            .map(|info| info.is_recently_connected())
            .unwrap_or_default()
```

shared/src/wg.rs
```diff
@@ -164,3 +165,94 @@ pub fn add_route(interface: &InterfaceName, cidr: IpNetwork) -> Result<bool, io:
...
+}
+impl PeerInfoExt for PeerInfo {
+    fn is_recently_connected(&self) -> bool {
+        const REJECT_AFTER_TIME: u64 = 180;
+
+        let last_handshake = self
+            .stats
+            .last_handshake_time
+            .and_then(|t| t.elapsed().ok())
+            .unwrap_or_else(|| SystemTime::UNIX_EPOCH.elapsed().unwrap());
```
strohel, line 254:
```suggestion
            .unwrap_or(Duration::MAX);
```

strohel (commented):
Looks good, I was finally able to grasp this in its entirety! Only the netlink response parsing code is somewhat gnarly, but that's likely due to inherent complexity of the netlink API itself.

I've added a few nits/suggestions.

shared/src/wg.rs
```diff
@@ -164,3 +165,94 @@ pub fn add_route(interface: &InterfaceName, cidr: IpNetwork) -> Result<bool, io:
 
 #[cfg(target_os = "linux")]
 pub use super::netlink::add_route;
+
+// TODO(jake): implement get_local_addrs() for macOS
```
strohel, line 169: stale TODO?

shared/src/netlink.rs
```diff
@@ -127,3 +147,88 @@ pub fn add_route(interface: &InterfaceName, cidr: IpNetwork) -> Result<bool, io:
...
+        // Only select addresses for helpful links
+        .filter(|nlas| nlas.iter().any(|nla| matches!(nla, address::nlas::Nla::Label(label) if links.contains(label))))
+        .filter_map(|nlas| nlas.iter().find_map(|nla| match nla {
+            // TODO(jake): support IPv6 addresses as well.
+            address::nlas::Nla::Address(name) if name.len() == 4 => {
+                let mut addr = [0u8; 4];
+                addr.copy_from_slice(name);
+                Some(IpAddr::V4(addr.into()))
+            },
+            address::nlas::Nla::Address(name) if name.len() == 16 => {
```
strohel, line 214: dtto stale TODO above?

server/src/api/user.rs
```diff
@@ -115,14 +124,25 @@ mod handlers {
...
+        session: Session,
+    ) -> Result<Response<Body>, ServerError> {
+        let conn = session.context.db.lock();
+        let mut selected_peer = DatabasePeer::get(&conn, session.peer.id)?;
+        selected_peer.update(
+            &conn,
+            PeerContents {
+                candidates: contents,
+                ..selected_peer.contents.clone()
+            },
```
strohel, line 139: May it happen that the client sends its "main peer address" that we already have as one of their candidates? Shall we filter it out? Shall we check in client and/or in server that candidates are unique?

client/src/main.rs
```diff
@@ -566,10 +515,42 @@ fn fetch(
...
+    let candidates = wg::get_local_addrs()?
+        .into_iter()
+        .map(|addr| SocketAddr::from((addr, device.listen_port.unwrap_or(51820))).into())
+        .collect::<Vec<Endpoint>>();
+    log::debug!("candidates: {:?}", candidates);
+    log::info!(
+        "reporting {} network interface addresses as ICE candidates...",
+        candidates.len()
+    );
+    log::debug!("viable ICE candidates: {:?}", candidates);
```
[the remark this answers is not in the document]
mcginty: Given that each peer will be running fetches at random times, I don't think this should be much of an issue, although I see what you're saying. I set it up this way so as to not make the existing behavior of innernet any slower than previous versions, but I think at the least I can move that `PUT /user/candidates` call before spending the larger amount of time on stepping through the candidates. I'll move it around :).

server/src/api/user.rs
mcginty: Yeah, that could certainly happen and could waste a 5 second step interval if it couldn't connect the first time. I'll add deduping and also just a total limit to the number of candidates.

shared/src/netlink.rs
mcginty: Ah yep, it sould be `?`.

shared/src/netlink.rs
mcginty: That's probably most correct, I'll do that.

mcginty: It should be in the right order already unless I'm misreading

shared/src/netlink.rs
mcginty: thanks :)

shared/src/types.rs
mcginty: I like yours better.

shared/src/wg.rs
mcginty: nice!

client/src/main.rs
```diff
@@ -566,10 +515,46 @@ fn fetch(
...
+    loop {
+        nat_traverse.step()?;
+        if nat_traverse.is_finished() {
+            break;
+        }
+        log::info!(
+            "{} unconnected peers remaining...",
+            nat_traverse.remaining()
+        );
+    }
```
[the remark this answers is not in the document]
mcginty: Hmm, there might be some situations where that would lead to an eventual connection if a peer is genuinely offline rather than behind a strict NAT.

I'll implement it just since it feels cleaner that way either way.

pri11er:
I was just giving this a try, going from 1.4.1, and server fails to start:
innernet-server[10422]: INFO  innernet_server::db > migrated db version from 1 to 2
innernet-server[10422]: Error: internal database error
innernet-server[10422]: Caused by:
innernet-server[10422]:       Invalid column type Null at index: 10, name: candidates
.service: Main process exited, code=exited, status=1/FAILURE

Revert and it starts ok again:
innernet-server[10561]:  INFO  innernet_server::db > migrated db version from 2 to 1
innernet-server[10561]:  INFO  innernet_server     > bringing up interface.
innernet-server[10561]:  INFO  innernet_server     > 12 peers added to wireguard interface.
innernet-server[10561]:  INFO  innernet_server     > innernet-server 1.4.1 starting.

mcginty:
You're on top of things @pri11er :).

This is fixed in #136.

PR #149, opened 2021-09-11
Add "wireguard" as a recommended dependency for Debian.

tommie:
It's very likely a user will want at least wireguard-dkms, and having
the userspace tools might be useful in an emergency. This metapackage
draws in both.

For automated installations in e.g. containers, use

    apt install --no-install-recommends

to avoid installing recommended packages.

I set up an auto-update APT repository on GitHub Pages for your deb files: https://github.com/tommie/innernet-debian


strohel (approved):
This makes good sense to me.

mcginty:
Makes sense to me too! Thanks @tommie.

PR #151, opened 2021-09-14
server: candidate reporting

mcginty:
Before, only clients would report local addresses for NAT traversal. Servers should too! This will be helpful in common situations when the server is run inside the same LAN as other peers, and there's no NAT hairpinning enabled (or possible) on the router.


strohel (commented):
Looks good in a quick retro-review.

PR #153, opened 2021-09-15
Clippy all targets

strohel:
Fix clippy lints in tests

CI: let clippy lint also tests

They deserve equal clippy love, right? :)


mcginty:
nice! thanks!

PR #155, opened 2021-09-18
client: fix minor typo ("exhuasted")

tianon:
Noticed this while I was reading through #134 trying to understand the overall logic. :innocent:


mcginty:
Ha! Nice catch, thanks :).

PR #162, opened 2021-10-01
feat: add INNERNET_NAT_EXCLUDE_CIDR env variable

pzeinlinger:
This comes in handy when nodes can talk to each other on multiple interfaces (eg. in a k8s cluster), but only some of them should be used for the innernet connection.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -173,6 +173,7 @@ dependencies = [
  "ipnetwork",
  "lazy_static",
  "log",
+ "netaddr2",
  "regex",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


mcginty:
Hey @pzeinlinger sorry for the AFKness. I'm in favor of something like this, but I think it fits better as a command-line argument since that's where similar features sit. Also, we have the `ipnetwork` crate for CIDR math, so no need to import `netaddr2` here :).

pzeinlinger:
This was just an idea and meant as a quick workaround. 
I don't work with Rust regularly, so if you want to implement this as a command-line argument (possibly allowing to exclude multiple cidrs from consideration) please go for it. :-)

PR #172, opened 2021-11-15
Make config/data directories configurable

### LINKED ISSUE #171: Configurable `SERVER_DATABASE_DIR` (or all directories in general)
I'm using `innernet-server` on OpenWRT, and it works great. However, by default, OpenWRT uses `tmpfs` for `/var` directory, and everything gets wiped out during restarts or upgrades. That includes innernet sqlite database.

Currently I'm using custom compilation, with paths altered, but is it possible, to make those directories configurable (for example via commandline flag)?

mcginty:
Fixes #171

### DIFF
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -18,8 +18,6 @@ pub mod wg;
 pub use types::*;
 
 lazy_static! {
-    pub static ref SERVER_CONFIG_DIR: &'static Path = Path::new("/etc/innernet-server");
-    pub static ref SERVER_DATABASE_DIR: &'static Path = Path::new("/var/lib/innernet-server");
     pub static ref REDEEM_TRANSITION_WAIT: Duration = Duration::from_secs(5);
 }
 
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -18,8 +18,6 @@ pub mod wg;
 pub use types::*;
 
 lazy_static! {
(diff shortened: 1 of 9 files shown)


shared/src/lib.rs
```diff
@@ -18,10 +18,6 @@ pub mod wg;
 pub use types::*;
 
 lazy_static! {
-    pub static ref CLIENT_CONFIG_DIR: &'static Path = Path::new("/etc/innernet");
-    pub static ref CLIENT_DATA_DIR: &'static Path = Path::new("/var/lib/innernet");
-    pub static ref SERVER_CONFIG_DIR: &'static Path = Path::new("/etc/innernet-server");
-    pub static ref SERVER_DATABASE_DIR: &'static Path = Path::new("/var/lib/innernet-server");
     pub static ref REDEEM_TRANSITION_WAIT: Duration = Duration::from_secs(5);
 }
```
strohel, line 22: Optional, because this was not introduced by this PR: `REDEEM_TRANSITION_WAIT` may be made just a `const REDEEM_TRANSITION_WAIT: Duration = Duration::from_secs(5);` without the `lazy_static`?
https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs is const from rustc 1.32.

strohel (approved):
Looks good, one little optional cleanup suggestion that was made possible by this PR.

shared/src/lib.rs
mcginty: nice catch! will fix :)

PR #177, opened 2021-12-03
(linux) wireguard-control: migrate from `wireguard-control-sys` to `netlink` crates

mcginty:
Expedited this change because of https://github.com/tonarino/innernet/issues/174, and it was good timing with the work done in the rust netlink project to support WireGuard's protocol.

Associated PR: https://github.com/little-dude/netlink/pull/210

PR #178, opened 2021-12-06
Run rustfmt on CI

strohel:
#177 fixes the formatting that fails this build.


mcginty (approved):
thanks! good to add this :).

PR #186, opened 2022-01-10
hostsfile: safely overwrite hosts file instead of clobbering

### LINKED ISSUE #183: hostsfile editing may wipe /etc/hosts when there is no space left on device
It happened that we ran out of available space on the root partition (such things happen embarrassingly often) on a host with innernet running.

The `innernet` process exited with an error, and (very likely) wiped the `/etc/hosts` file:
```bash
[me@buildripper ~]$ sudo systemctl status innernet@tonari
× innernet@tonari.service - innernet client daemon for tonari
     Loaded: loaded (/etc/systemd/system/innernet@tonari.service; enabled; vendor preset: disabled)
     Active: failed (Result: exit-code) since Mon 2022-01-03 00:36:48 UTC; 1 day 21h ago
    Process: 3960906 ExecStart=/usr/bin/innernet up tonari --daemon --interval 60 (code=exited, status=1/FAILURE)
   Main PID: 3960906 (code=exited, status=1/FAILURE)
        CPU: 4ms

Jan 03 00:36:48 buildripper innernet[3960906]: [*] fetching state from server.
Jan 03 00:36:48 buildripper innernet[3960906]: [*] peers are already up to date.
Jan 03 00:36:48 buildripper innernet[3960906]: [E] tonari - No space left on device (os error 28)
Jan 03 00:36:48 buildripper systemd[1]: innernet@tonari.service: Main process exited, code=exited, status=1/FAILURE
Jan 03 00:36:48 buildripper systemd[1]: innernet@tonari.service: Failed with result 'exit-code'.
Jan 03 00:36:48 buildripper systemd[1]: innernet@tonari.service: Scheduled restart job, restart counter is at 16.
Jan 03 00:36:48 buildripper systemd[1]: Stopped innernet client daemon for tonari.
Jan 03 00:36:48 buildripper systemd[1]: innernet@tonari.service: Start request repeated too quickly.
Jan 03 00:36:48 buildripper systemd[1]: innernet@tonari.service: Failed with result 'exit-code'.
Jan 03 00:36:48 buildripper systemd[1]: Failed to start innernet client daemon for tonari.
[me@buildripper ~]$ file /etc/hosts
/etc/hosts: empty
```

It seems the culprit is non-defensive writing of the hostsfile in https://github.com/tonarino/innernet/blob/6be3e61074b6762c88d13a292b21cd78e681418f/hostsfile/src/lib.rs#L221-L227

It would be better if innernet:
- had written a temporary file like `/etc/hosts.random84937493` and then renamed that to `/etc/hosts` (which is atomic).
- issued a warning that hostsfile could have not been updated instead of exiting.

mcginty:
In this case, I just opted to use a millisecond timestamp instead of a random number to avoid extra dependencies.

fixes #183

### DIFF
--- a/hostsfile/src/lib.rs
+++ b/hostsfile/src/lib.rs
@@ -253,8 +253,14 @@ impl HostsBuilder {
         }
 
         match Self::write_and_swap(&temp_path, &hosts_path, &s) {
-            Err(_) => Self::write_clobber(&hosts_path, &s),
-            _ => Ok(()),
+            Err(_) => {
+                log::debug!("wrote hosts file with the clobber fallback strategy");
+                Self::write_clobber(&hosts_path, &s)
+            },
+            _ => {
+                log::debug!("wrote hosts file with the write-and-swap strategy");
+                Ok(())
+            }
         }
     }
 
--- a/hostsfile/src/lib.rs
+++ b/hostsfile/src/lib.rs
@@ -252,16 +252,34 @@ impl HostsBuilder {
             writeln!(&mut s, "{}", line)?;
         }
 
+        match Self::write_and_swap(&temp_path, &hosts_path, &s) {
+            Err(_) => Self::write_clobber(&hosts_path, &s),
+            _ => Ok(()),
+        }
+    }
+
+    fn write_and_swap(temp_path: &Path, hosts_path: &Path, contents: &[u8]) -> io::Result<()> {
+        std::fs::copy(&hosts_path, &temp_path)?;
+        OpenOptions::new()
+            .create(false)
+            .read(true)
+            .write(true)
+            .truncate(true)
+            .open(&temp_path)?
+            .write_all(&contents)?;
+
+        std::fs::rename(temp_path, hosts_path)?;
+        Ok(())
+    }
+
+    fn write_clobber(hosts_path: &Path, contents: &[u8]) -> io::Result<()> {
         OpenOptions::new()
             .create(true)
             .read(true)
             .write(true)
             .truncate(true)
-            .open(&temp_path)?
-            .write_all(&s)?;
-
-        std::fs::rename(temp_path, hosts_path)?;
-
+            .open(hosts_path)?
+            .write_all(&contents)?;
         Ok(())
     }
 }
--- a/hostsfile/src/lib.rs
+++ b/hostsfile/src/lib.rs
@@ -265,6 +265,7 @@ impl HostsBuilder {
     }
 
     fn write_and_swap(temp_path: &Path, hosts_path: &Path, contents: &[u8]) -> io::Result<()> {
+        // Copy the file we plan on modifying so its permissions and metadata are preserved.
         std::fs::copy(&hosts_path, &temp_path)?;
         Self::write_clobber(temp_path, contents)?;
         std::fs::rename(temp_path, hosts_path)?;
--- a/hostsfile/src/lib.rs
+++ b/hostsfile/src/lib.rs
@@ -5,7 +5,7 @@ use std::{
     io::{self, BufRead, BufReader, ErrorKind, Write},
     net::IpAddr,
     path::{Path, PathBuf},
-    result,
+    result, time::{SystemTime, UNIX_EPOCH},
 };
 
 pub type Result<T> = result::Result<T, Box<dyn std::error::Error>>;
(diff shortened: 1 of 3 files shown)


strohel (approved):
Looks good, thanks for looking into this!

One consideration that came to my mind only now: Assuming the innernet running as ordinary user mode: previously only writable /etc/hosts was needed, but now it may require the `/etc/` directory itself to be writable? (for the creation of the temp file, not sure what permissions are checked for the rename)

I don't have any idea how to avoid this while still being atomic -- I don't think this is a blocker for merging this.

mcginty:
Yeah, I think that *might* have been why we did it the other way, although my memory doesn't serve particularly well (a good reason to add comments now...).

We could also have logic that tries to do this renaming method *first* and then falls back to the less safe truncate-and-rewrite option if the creation of a new file doesn't work. Do you think that's too much? I'll also look into seeing how easy it is to check available space on the mount and failing *before* truncating if that's possible...

mcginty:
Converting to a draft since this needs more thought before merging.

Actually, I do remember that originally `hostsfiles` used to do the safer rename-style method: https://github.com/tonarino/innernet/commit/449b4b8278496634cae9cb87fc3dc20e71222f7b

...and then I changed it to support running the innernet client in non-root environments more easily.

Here are two ideas for new logic, depending on our tolerance for complexity:
## complicated fallback logic
1. Attempt to do the `copy-write-replace` within the parent folder
2. If that fails, check [`statvfs`](https://docs.rs/nix/latest/nix/sys/statvfs/struct.Statvfs.html) for available blocks on the filesystem and write-in-place if it seems likely to succeed and hope for the best

## simpler write-in-place logic
Just do the second step from the complicated option and hope for the best

strohel:
I've timeouted trying to come up with something smart, so...

I think _complicated fallback logic_ would be better. The step 1. will catch all users running innernet as root, which may be the (vast?) majority. We could optionally warn (perhaps once per program lifetime) if we fallback to 2. so that non-root users know their tradeoffs.

mcginty:
pro-tip: never accidentally `git reset main`, then `git push -f`...

mcginty:
@strohel implemented the swap-and-replace with clobber fallback (opted not to add code to check `statvfs` for now) - feel free to take a peek if you have time, otherwise i'll merge soon :).

hostsfile/src/lib.rs
```diff
@@ -218,14 +252,80 @@ impl HostsBuilder {
...
             .create(true)
             .read(true)
             .write(true)
             .truncate(true)
             .open(hosts_path)?
-            .write_all(&s)?;
-
+            .write_all(contents)?;
         Ok(())
     }
```
strohel, line 290: Do we actually need to `std::fs::copy(&hosts_path, &temp_path)` in `write_and_swap()`? If I'm not mistaken we've already _read_ hosts file up in `write_to()`.

We should be able to call `write_clobber(temp_path, contents)` in `write_and_swap()`.

Alternatively, if we want to prevent a collision of temporary file names, we could use different more defensive OpenOptions in the two:
write_and_swap(): `OpenOptions::new().`**`create_new()`**`.write(true)`
write_clobber(): `OpenOptions::new().create(true).write(true).truncate(true)`

strohel (commented):
Looks good! Makes sense to start with minimal reasonable changes - this would have already prevented the problem we experienced.

I've added just one technicality comment.

hostsfile/src/lib.rs
mcginty: I added the `std::fs::copy(&hosts_path, &temp_path)` back in since it was in a previous version of `hostsfiles` as a simple way to preserve permissions from the file that it would be overwriting. Seemed reasonable, especially since hostsfiles tend to be small and CoW filesystems are getting more prevalent anyway :).

Good point about using `write_clobber` within `write_and_swap` - done!

strohel: > I added the `std::fs::copy(&hosts_path, &temp_path)` back in since it was in a previous version of `hostsfiles` as a simple way to preserve permissions from the file that it would be overwriting.

Oh, that's a nice trick! We could optionally add a comment to explain, but that's a super-nit.

strohel (approved):
Looks great!

hostsfile/src/lib.rs
mcginty: Will do! You're right, that line definitely deserves an explanation.

Issue #191, opened 2022-01-31
IPv6 NAT candidate reporting


Enter-tainer:
This is part of the output when I run `innernet show` on my device. These devices(arch, router, home-server) all have a ipv4 address (a private IP address, like 192.168.1.1), and a ipv6 address(a global unicast address). But it seems that only the ipv4 one is being used in innernet. If the ipv6 address can be used to connect peers, we can simply get rid of annoying NATs, and connect to peers directly through ipv6. 

So is it possible to use ipv6 addresses in innernet?

I've been using innernet for a while, it is really a great project! Many thanks for your great efforts.
```
 peer: arch (2t9yDWwyTY...)
    ip: 10.42.1.1
    endpoint: 192.168.147.80:60688
    last handshake: 19 minutes, 18 seconds ago
    transfer: 3.08 MiB received, 7.34 MiB sent
  peer: router (hLu6TevaY5...)
    ip: 10.42.1.3
    endpoint: 192.168.1.1:55552
    transfer: 0 B received, 1418.57 KiB sent
  peer: home-server (VhItncIvyG...)
    ip: 10.42.1.4
    endpoint: 192.168.1.7:33498
    last handshake: 1 minute, 43 seconds ago
    transfer: 36.27 KiB received, 133.89 KiB sent
```

mcginty:
innernet can use IPv6 endpoints, and there are a couple reasons these endpoints aren't being used right now in your setup:

1) Your coordinating server is probably setup with an IPv4 endpoint, so clients are connecting to the server via IPv4 and the server is passing those addresses it sees to fellow peers. If the server were being reached over IPv6 it would be reporting other peers' IPv6 addresses.
2) Right now the innernet client is only reporting IPv4 addresses as NAT candidates, as it's a relatively new addition. This issue is a good reminder that it's time to add IPv6 addresses as other candidates though!

The quickest interim solution if your IPv6 addresses are not temporary would be to use `innernet override-endpoint` to manually specify an IPv6 endpoint that peers should use to connect.

mcginty:
I went ahead and changed the title so that I can use it as the issue for adding IPv6 NAT candidates :). Thanks again for opening the issue!

Enter-tainer:
I'm trying to implement this. I found that the client will collect its ip addresses, and then upload them to the server.
```rs
    let candidates: Vec<Endpoint> = get_local_addrs()?
        .filter(|ip| !nat.is_excluded(*ip))
[...]

PR #192, opened 2022-01-31
enable IPv6 NAT candidate reporting

mcginty:
This comes from the conversation in #191

**previous behavior**: filter out all IPv6 addresses from being reported as NAT candidates.

**new behavior**: Begin reporting IPv6 addresses that are *global unicast* scope only.

An argument could be made for also reporting [unique local addresses](https://en.wikipedia.org/wiki/Unique_local_address) as well, but I think it's less important and more likely to increase noise with little benefit to getting through any different types of NATs.

I'd love to hear from people more experienced with managing IPv6 addresses if that's not an appropriate strategy.

### DIFF
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -2,7 +2,7 @@ pub use anyhow::Error;
 use std::{
     fs::{self, File, Permissions},
     io,
-    net::Ipv6Addr,
+    net::{IpAddr, Ipv6Addr},
     os::unix::fs::PermissionsExt,
     path::Path,
     time::Duration,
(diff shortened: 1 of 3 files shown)


shared/src/lib.rs
```diff
@@ -97,5 +98,21 @@ pub fn _get_local_addrs() -> Result<impl Iterator<Item = std::net::IpAddr>, io::
...
+            || ((ip.segments()[0] == 0x2001) && (ip.segments()[1] == 0xdb8)) // documentation
+            || (ip.segments()[0] & 0xffc0) == 0xfe80 // unicast link local
+            || (ip.segments()[0] & 0xfe00) == 0xfc00) // unicast local
+    }
+
+    Ok(_get_local_addrs()?
+        .filter(|ip| {
+            ip.is_ipv4()
+                || matches!(ip,
+            IpAddr::V6(v6) if is_unicast_global(v6))
```
strohel, line 115: This is a nice trick to concisely check something on an enum variant, I need to remember to use it, too.

(if only `rustfmt` didn't spoil it by being somewhat clueless with this expression)

strohel (commented):
Quick post-mortem review: both intended functionality and code look good to me.

PR #193, opened 2022-01-31
meta: switch from ipnetwork to ipnet

mcginty:
[`ipnet`](https://github.com/krisprice/ipnet) seems to be what the world has settled on as the de facto CIDR crate.

### DIFF
--- a/server/src/db/cidr.rs
+++ b/server/src/db/cidr.rs
@@ -1,7 +1,7 @@
 use crate::ServerError;
-use ipnetwork::IpNetwork;
+use ipnet::IpNet;
 use rusqlite::{params, Connection};
-use shared::{Cidr, CidrContents};
+use shared::{Cidr, CidrContents, IpNetExt};
 use std::ops::Deref;
 
 pub static CREATE_TABLE_SQL: &str = "CREATE TABLE cidrs (
@@ -56,8 +56,8 @@ impl DatabaseCidr {
 
             let closest_parent = cidrs
                 .iter()
-                .filter(|current| cidr.is_subnet_of(current.cidr))
-                .max_by_key(|current| current.cidr.prefix());
+                .filter(|current| current.cidr.contains(cidr))
+                .max_by_key(|current| current.cidr.prefix_len());
 
             if let Some(closest_parent) = closest_parent {
                 if closest_parent.id != *parent_id {
@@ -70,7 +70,7 @@ impl DatabaseCidr {
             }
 
             let parent_cidr = Self::get(conn, *parent_id)?.cidr;
-            if !parent_cidr.contains(cidr.network()) || !parent_cidr.contains(cidr.broadcast()) {
+            if !parent_cidr.contains(&cidr.network()) || !parent_cidr.contains(&cidr.broadcast()) {
                 log::warn!("tried to add a CIDR with a network range outside of its parent.");
                 return Err(ServerError::InvalidQuery);
             }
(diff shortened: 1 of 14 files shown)


server/src/db/cidr.rs
```diff
@@ -109,14 +114,12 @@ impl DatabaseCidr {
...
-        let cidr = IpNetwork::new(
-            ip.parse()
-                .map_err(|_| rusqlite::Error::ExecuteReturnedResults)?,
-            prefix,
-        )
-        .map_err(|_| rusqlite::Error::ExecuteReturnedResults)?;
+        let ip = ip_str
+            .parse()
+            .map_err(|_| rusqlite::Error::ExecuteReturnedResults)?;
+        let cidr = IpNet::new(ip, prefix).map_err(|_| rusqlite::Error::ExecuteReturnedResults)?;
```
strohel, line 122: (preexisting)
I see that our hands are tied by https://docs.rs/rusqlite/latest/rusqlite/enum.Error.html

But it seems that there are some variants that would hopefully not be worse and would allow forwarding underlying error? I'm looking at `FromSqlConversionFailure(usize, Type, Box<dyn Error + Send + Sync + 'static>),` or `UserFunctionError(Box<dyn Error + Send + Sync + 'static>),`.

It is possible thought that visibility may not allow us to create these.

shared/src/lib.rs
```diff
@@ -116,3 +117,29 @@ pub fn get_local_addrs() -> Result<impl Iterator<Item = std::net::IpAddr>, io::E
...
+    }
+
+    fn is_assignable(&self, ip: &IpAddr) -> bool {
+        self.contains(ip)
+            && match self {
+                IpNet::V4(_) => {
+                    self.prefix_len() >= 31 || (ip != &self.network() && ip != &self.broadcast())
+                },
+                IpNet::V6(_) => self.prefix_len() >= 127 || ip != &self.network(),
+            }
```
strohel, line 143: Curiosity question: why do we need to make exception for the "full prefix" and "full minus one prefix"? Without this exception, such CIDRs would not be assignable at all in IPv4 case and very limited in IPv6 case - perhaps that's the reason?

shared/src/wg.rs
```diff
@@ -32,17 +32,17 @@ fn cmd(bin: &str, args: &[&str]) -> Result<std::process::Output, io::Error> {
 }
 
 #[cfg(target_os = "macos")]
-pub fn set_addr(interface: &InterfaceName, addr: IpNetwork) -> Result<(), io::Error> {
+pub fn set_addr(interface: &InterfaceName, addr: IpNet) -> Result<(), io::Error> {
     let real_interface = wireguard_control::backends::userspace::resolve_tun(interface)?;
 
-    if addr.is_ipv4() {
+    if matches!(addr, IpNet::V4(_)) {
```
strohel, line 38: Optional: this would be also a candidate for `IpNetExt` convenience method.

strohel (commented):
Nicely straightforward port! I've added some curiosity questions, most of them on preexisting stuff.

server/src/db/cidr.rs
mcginty: Good point, I'll use FromSqlConversionFailure instead.

shared/src/lib.rs
mcginty: Yep, exactly. It's more likely an issue in the IPv4 space, but in a /31 the only two addresses in that CIDR would technically be the network and broadcast address, but there is an RFC that clarifies that /31 is special-cased in the same way that /32 is for that reason. I'll have to find the specific RFC that states this.

shared/src/wg.rs
mcginty: true!

shared/src/lib.rs
strohel: Oh I see, this is special-cased in the standard. Thanks!

PR #194, opened 2022-02-01
server: add better validation to the associations endpoint

mcginty:
Previously, the server didn't care too much about what the parameters for an association were - this adds more proper validation to the inputs to avoid any odd situations.

### DIFF
--- a/server/src/db/association.rs
+++ b/server/src/db/association.rs
@@ -57,6 +57,30 @@ impl DatabaseAssociation {
             cidr_id_2,
         } = &contents;
 
+        // Verify an existing association doesn't currently exist
+        let existing_associations: usize = conn.query_row(
+            "SELECT COUNT(*)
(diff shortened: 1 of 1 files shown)


server/src/db/association.rs
```diff
@@ -89,3 +113,103 @@ impl DatabaseAssociation {
...
+        };
+        let res = server
+            .form_request(
+                test::ADMIN_PEER_IP,
+                "POST",
+                "/v1/admin/associations",
+                &contents,
+            )
+            .await;
+        assert!(!res.status().is_success());
```
strohel, line 190: Super-nit: using `.is_client_error()` as already done above could be a tiny bit more descriptive here.

strohel (commented):
Looks good, and thanks for the nice tests. I've added one supernit for the sake of exercise. ;-)

PR #195, opened 2022-02-01
client: re-attempt NAT traversal on all unconnected peers

### LINKED ISSUE #173: When a peer isn't modified in fetch, NAT traversal doesn't attempt to re-establish a broken connection
Thus, sometimes a NAT traversal that *should* be attempted when a handshake hasn't been completed in the last 3 minutes is not attempted.

mcginty:
fixes #173

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -616,7 +616,7 @@ fn fetch(
         Err(e) => return Err(e.into()),
         _ => {},
     }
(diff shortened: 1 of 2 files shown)


strohel (commented):
Makes sense to me. I've tested this locally, but the the condition to trigger this case doesn't seem to be common?

Issue #218, opened 2022-05-11
Add --yes flag to innernet-server related commands for automation


jon77p:
In commit ddac328, the `--yes` flag was made available to the `innernet` client binary, however the same flaw were not added to the `innernet-server` binary. This prevents automations (such as using ansible to manage an innernet deployment) from being made possible on commands like "uninstall" on the internet server.

The logic added in the referenced commit seems to be straightforward enough that changes to the `innernet-server` binary should be easy, and I probably will look into making a PR if I implement the changes myself, but I also wanted to track this here for visibility!

jon77p:
Thanks @mcginty for merging in PR #221! I'll mark this issue as resolved as it's been merged into main.

mcginty:
Thanks for the PR! Sorry it took so long to merge.

PR #221, opened 2022-05-13
Add missing --yes flag to innernet-server uninstall command

jon77p:
As detailed in issue #218, there currently is no way to automate `innernet-server` network uninstalls due to the `--yes` flag missing. In commit ddac328, the logic to add this flag and check if it is set/not was added to the `innernet` client program for all commands.

In the changes in this PR, I copied the same logic and structure additions from `client/src/main.rs` to the `Uninstall` command implementation, and also updated the `uninstall()` function parameters and conditional logic to match accordingly.

Additionally, I have tested to ensure that the change in this PR builds without errors and successfully uninstalls the network without a confirmation as expected when `--yes` is added.

If there's any additional questions/improvements needed, just let me know and I'd be happy to update this PR!

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -71,7 +71,13 @@ enum Command {
     },
 
     /// Permanently uninstall a created network, rendering it unusable. Use with care.
-    Uninstall { interface: Interface },
+    Uninstall {
+        interface: Interface,
+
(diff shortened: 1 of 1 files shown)


mcginty:
Thanks @jon77p! Looks good to me.

PR #222, opened 2022-05-21
External endpoint IP prompt change

richardjs:
This changes the prompt for the external endpoint IP when creating a network. Specifically, it:

- Changes the "public IP address" to "external endpoint IP address" (as far as I see, the term "public IP" isn't used elsewhere in innernet)
- Changes "auto-fill" to "auto-detect"
- Removes an (apparently?) superfluous print

Small changes! But the existing prompt briefly confused me, so I thought it was worth the suggestion.


mcginty:
Looks good! Thanks, I like your version better.

alerque (approved):

Issue #228, opened 2022-07-28
Segfault at the end of the `innernet-server new` wizard on Alpine Linux


hellerbarde:
Hi!

I am new to Innernet and wanted to try it out after reading your blogpost. 

I'm using an Alpine Linux VM (`x86_64-alpine-linux-musl`) that I already used as a Wireguard host. After a bit of getting started, I compiled the innernet-server binary with cargo and ran `innernet-server new`. It crashed with a segmentation fault. 

Then I compiled it again with debug flags and ran the valgrind memcheck thing (I'm not a valgrind expert, but GDB didn't show anything obvious)

Here's how I built and ran it:
```
git clone <this repository> && cd innernet
cargo build --bin innernet-server

valgrind --tool=memcheck target/debug/innernet-server new
```

Here's the output:

```
# valgrind --tool=memcheck target/debug/innernet-server new
==10010== Memcheck, a memory error detector
==10010== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==10010== Using Valgrind-3.17.0 and LibVEX; rerun with -h for copyright info
==10010== Command: target/debug/innernet-server new
==10010==

Time to setup your innernet network.

Your network name can be any hostname-valid string, i.e. "evilcorp", and
your network CIDR should be in the RFC1918 IPv4 (10/8, 172.16/12, or 192.168/16),
or RFC4193 IPv6 (fd00::/8) ranges.

The external endpoint specified is a <host>:<port> string that is the address clients
will connect to. It's up to you to forward/open ports in your routers/firewalls
as needed.

For more usage instructions, see https://github.com/tonarino/innernet#usage

==10010== Conditional jump or move depends on uninitialised value(s)
==10010==    at 0xB8CA55: mal0_clear (malloc.c:346)
==10010==    by 0xB8CA55: calloc (malloc.c:364)
==10010==    by 0x886B8B: alloc::alloc::alloc_zeroed (alloc.rs:160)
==10010==    by 0x886D1C: alloc::alloc::Global::alloc_impl (alloc.rs:171)
==10010==    by 0x88823C: <alloc::alloc::Global as core::alloc::Allocator>::allocate_zeroed (alloc.rs:236)
==10010==    by 0x898FB1: alloc::raw_vec::RawVec<T,A>::allocate_in (raw_vec.rs:186)
[...]

PR #229, opened 2022-07-29
Update sqlite dependencies + cargo update

strohel:
Let's see if that helps with #228. CC @hellerbarde.

### Fix clippy warnings

Especially https://rust-lang.github.io/rust-clippy/master/index.html#format_push_string was a bit
more involved, review requested.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,9 +24,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.57"
+version = "1.0.58"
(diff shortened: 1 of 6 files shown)


PabloMansanet (approved):
LGTM, it wasn't immediately clear that the `writeln` approach had the exact same semantics (more efficiently) than the push_str one, but it's well explained in the lint page so this looks safe to merge.

bschwind (commented):
Post-merge LGTM!

PR #230, opened 2022-07-29
avoid a rustc bug

aliemjay:
Hi,

https://github.com/rust-lang/rust/pull/98835 fixes a rustc bug that makes it incorrectly accepts the following code:
```rust
fn test<'a: 'a>() { // 'a is longer than the entire function body
    let f = |_: &'a str| {}; // expects an argument that lives as long as 'a
    f(&String::new()); // the passed reference is shorter that `'a`
}
```

The crater run found that `innernet` will break with this change because it *indirectly* relies on this bug. I'm not sure when this change reaches stable, if ever, but it's better to not rely on such a bug. This PR tries to fix this.

If you have an opinion on this change, please let me know here or  by commenting on the PR.

You can install and test the toolchain locally by running the command: `rustup-toolchain-install-master cb1f7c96119295882e08b09b4bd01051669c071b`

### DIFF
--- a/shared/src/types.rs
+++ b/shared/src/types.rs
@@ -259,11 +259,11 @@ impl<'a> CidrTree<'a> {
         }
     }
 
-    pub fn children(&self) -> impl Iterator<Item = CidrTree> {
+    pub fn children(&self) -> impl Iterator<Item = CidrTree<'_>> {
         self.cidrs
             .iter()
             .filter(move |c| c.parent == Some(self.contents.id))
-            .map(move |c| Self {
+            .map(move |c| CidrTree {
                 cidrs: self.cidrs,
                 contents: c,
             })


shared/src/types.rs
```diff
@@ -259,11 +259,11 @@ impl<'a> CidrTree<'a> {
         }
     }
 
-    pub fn children(&self) -> impl Iterator<Item = CidrTree> {
+    pub fn children(&self) -> impl Iterator<Item = CidrTree<'_>> {
         self.cidrs
             .iter()
             .filter(move |c| c.parent == Some(self.contents.id))
-            .map(move |c| Self {
+            .map(move |c| CidrTree {
```
aliemjay, line 266: The reason for this change is that `Self` (which is just a sugar for `CidrTree<'a>`) brings  the the lifetime `'a` to the scope of the returned opaque type. Use `CidrTree` instead to allow rustc to infer a shorter lifetime (no necessarily `'a`).

Generally, it's recommended to use the struct name instead of `Self` to allow rust to infer the appropriate lifetime rather than unnecessarily forcing a specific lifetime on the created

bschwind: Oh this is an interesting one! I've often used `Self` in places like this, good to know the kind of issues that can cause.

bschwind (approved):
This looks good to me. @mcginty @strohel what do you think?

strohel (approved):
This is an interesting one! It also didn't occur to me that saying `Self` could bring an unwanted lifetime bound in. Approving.

@aliemjay can you please rebase on `main` before we merge this? Or I can do that if I have force-push permissions to your branch. We've fixed clippy warnings in the mean time, so that should make the CI green.

aliemjay:
I'm on mobile now, so I can only use github ui to merge the branch, if this is not suitable, feel free to close this and commit the change yourself.

strohel:
@aliemjay oh that works, too merging this will rebase the merge away. Thanks!

Issue #236, opened 2022-11-06
Can't find debian packages


rehanog:
Apologies if I am missing something obvious.

From the README:
Fetch the appropriate .deb packages from https://github.com/tonarino/innernet/releases

I can't for the life of me find any packages on that releases page. Only see the sources

Help appreciated,

Many thanks

Rehan

bschwind:
Related issue: #227

We need to get back in the habit of generating .deb files on each release, or perhaps removing that line from the README...

It's fairly simple to generate the debian files yourself, but let me know if you run into issues.

strohel:
Hi @rehanog, sorry for incorrect info in readme. We've fixed that in https://github.com/tonarino/innernet/pull/238. Research on external build systems continues in https://github.com/tonarino/innernet/issues/203.

Issue #227, opened 2022-07-19
Documentation says Ubuntu/Debian .DEB files are the Release section 


bmullan:
Documentation says...

> Innernet consists of two binaries: innernet-server, the coordinating server, and innernet (also aliased as inn for easy access), the client that peers run to access the network.
> 
> First, let's [install ](https://github.com/tonarino/innernet#installation)[innernet-server](https://github.com/tonarino/innernet#installation) on a server we have sitting around, create a new network, and start it up:
> 
> First, let's install innernet-server on a server we have sitting around, create a new network, and start it up ...

But I do not see any releases w .DEB files just source and tar.gz **_except in the BETA releases_**

Do the .DEB files exist and if so where are they...

thanks
brian

bschwind:
Hi @bmullan , 

See the issue [here](https://github.com/tonarino/innernet/issues/203#issuecomment-1069520125) for the current status on *.deb files.

We should fix up the documentation for sure.

You can generate the deb files on the machine you plan to use them with:

```bash
cargo install cargo-deb # if it isn't installed already
cargo deb -p client
cargo deb -p server
```

Assuming you have `cargo` installed. If you don't, you can install it from [rustup.rs](https://rustup.rs/)

bschwind:
Closing as a duplicate of #212, but feel free to comment here if you need any help generating the files.

Issue #212, opened 2022-04-24
Latest release (1.5.4) doesn't seem to link the deb files


natcl:
Is this normal ?

Thanks !

prouveyrol:
Hello.
See https://github.com/tonarino/innernet/issues/203

strohel:
Hi @natcl, sorry for incorrect info in readme. We've fixed that in #238. Research on external build systems continue in #203.

Issue #203, opened 2022-03-16
Binary packages no longer built (was: 1.5.4 package doesn't work on Debian buster)


bensteinberg:
This isn't really a problem, just a report -- when I upgraded a mixed set of buster and bullseye machines to 1.5.4, I found that  the service wouldn't start on the buster machines. The error in the syslog was

    innernet[28207]: /usr/bin/innernet: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by /usr/bin/innernet)

so I downgraded the buster machines to innernet 1.5.3. I'm not sure whether this has to do with the OS of the underlying system that GitHub workflow uses to build the Debian packages. My buster machines have libc6 version 2.28-10, and the bullseye machines have 2.31-13+deb11u2.

alerque:
This project should not be releasing deb files ... there are lots of different variants of Ubuntu systems (Debian, Ubuntu, others, different releases, etc). Packaging info should be tracked in a separate packaging repository and binary packages released to wherever each distro needs them (the main upstream distro if possible, PPAs or other user repositories otherwise).

mcginty:
I'm in agreement, I think innernet should stop releasing "official" deb or rpm files since they're bound to continue to cause headaches (or overload people with the variations necessary for it to work for all users).

That said, @bensteinberg, just run:

```bash
cargo install cargo-deb # if it isn't installed already
cargo deb -p client
```

and you'll have a working deb linked and built correctly for your OS and architecture :).

mcginty:
@bensteinberg Don't worry, I don't feel hassled in the slightest! This is quite helpful to talk through.

Yeah, I agree that innernet currently fits best as a PPA or external apt repository vs. being packaged, given its rate of change still. As for build machines, I can supply that either via my own machines or configuring GitHub actions to generate the right artifacts, the main concern is just hosting them in a convenient place and keeping them maintained :).

Thanks for reminding me about homebrew! It's not currently automated, and I forgot to update it. I think now that innernet is in more wide-spread usage, I'm going to make a PR to submit innernet to the official Homebrew repository.

alerque:
> I'm going to make a PR to submit innernet to the official Homebrew repository.

Definitely recommended. Besides making it easy to get started with for end users, they have pretty good CI checks on their main repository and it allows anybody submitting a PR to bump on updates so you don't have to feel on the line for it.

tommie:
[...]

PR #238, opened 2022-12-01
README: clarify that we're no longer building .deb and RPM

strohel:
...and say that we're looking for help.

Relates to #203 (main issue), #197, #202, #212, #227, #236, #237.


bschwind (approved):
LGTM!

PR #239, opened 2022-12-01
CI: use stable clippy

strohel:
Using nightly one leads to having to resolve lints more frequently than rust releases, which is getting inconvenient.

It also leads to a small compatibility issue with stable clippy: if we need to #[allow] a lint that is only present in nightly clippy, then stable clippy complains about unknown lint in the `allow`. :shrug:


bschwind (approved):
LGTM!

PR #243, opened 2023-01-12
Fix rand_core dependency, cargo update

### LINKED ISSUE #242: `cargo install` fails with unresolve import of `rand_core::OsRng`
`cargo install --locked` works, so it's likely that `rand_core` did a patch version upgrade that broke its API? Note that `cargo install` without `--locked` won't use the `Cargo.lock` file.

```
$ cd innernet
$ cargo install --path client

error[E0432]: unresolved import `rand_core::OsRng`
  --> wireguard-control/src/key.rs:35:25
   |
35 |         use rand_core::{OsRng, RngCore};
   |                         ^^^^^ no `OsRng` in the root

error[E0432]: unresolved import `rand_core::OsRng`
  --> wireguard-control/src/key.rs:50:25
   |
50 |         use rand_core::{OsRng, RngCore};
   |                         ^^^^^ no `OsRng` in the root

warning: unused import: `RngCore`
  --> wireguard-control/src/key.rs:50:32
   |
50 |         use rand_core::{OsRng, RngCore};
   |                                ^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `RngCore`
  --> wireguard-control/src/key.rs:35:32
   |
35 |         use rand_core::{OsRng, RngCore};
   |
```

strohel:
Should fix #242.

We depend on `getrandom` feature of `rand_core`, but we weren't specifying that.

It worked because `curve25519-dalek 4.0.0-pre.2` was pulling the `getrandom` feature of `rand_core`, but after updating it to `4.0.0-pre.5` it stopped pulling `rand_core` altogether.

After the cargo-update, all out dependencies are on their newest version, sans clap 3.2, which could be updated to 4.0. It has quite some breaking API changes, so I'll leave that for a separate PR.

This should also fix https://github.com/tonarino/innernet/security/dependabot/1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,18 +15,18 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.19"
+version = "0.7.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
+checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
 dependencies = [
  "memchr",
 ]
 
 [[package]]
 name = "anyhow"
-version = "1.0.65"
+version = "1.0.68"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "98161a4e3e2184da77bb14f02184cdd111e83bbbcc9979dfee3c44b9a85f5602"
+checksum = "2cb2f989d18dd141ab8ae82f64d1a8cdd37e0840f73a406896cf5e99502fab61"
 
 [[package]]
 name = "atty"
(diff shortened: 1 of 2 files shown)


skywhale (approved):
LGTM

Thank you for the quick fix.

I was wondering after posting my message, because `Cargo.lock` had 0.6.4 of `rand_core`, which is the lastest. I didn't know transitive dependencies could behave this way. Good catch @strohel :)

strohel:
Yeah `feature` dependencies are somewhat less visible than version dependencies - they are not mentioned in `Cargo.lock`. And there's similar caveat when using workspaces: for each dependency a union of pulled features is made, hence it is easy to forget specifying all needed features when adding a workspace crate. It also means that `cargo build -p audio-playground` can produce slightly different result than `cd crates/audio-playground && cargo build` :shrug:.

bschwind (approved):
LGTM!

mcginty (approved):

Issue #214, opened 2022-04-27
add name flag to disable-peer command


maaft:
It would be great to add a `--name` flag to the `disable-peer` sub-command. Now its cumbersome to disable-peers automatically i.e. without human console input.

PR #248, opened 2023-02-22
Add CLI parameters for disable/enable peer

linuskendall:
Adds a --name parameter and --yes parameter to the enable-peer/disable-peer. Helps with large network management where we need some automated tools to disable/enable peers.

Fixes tonarino/innernet#214.

### DIFF
--- a/shared/src/prompts.rs
+++ b/shared/src/prompts.rs
@@ -1,8 +1,8 @@
 use crate::{
     interface_config::{InterfaceConfig, InterfaceInfo, ServerInfo},
     AddCidrOpts, AddDeleteAssociationOpts, AddPeerOpts, Association, Cidr, CidrContents, CidrTree,
-    DeleteCidrOpts, Endpoint, Error, Hostname, IpNetExt, ListenPortOpts, OverrideEndpointOpts,
-    Peer, PeerContents, RenamePeerOpts, EnableDisablePeerOpts, PERSISTENT_KEEPALIVE_INTERVAL_SECS,
+    DeleteCidrOpts, EnableDisablePeerOpts, Endpoint, Error, Hostname, IpNetExt, ListenPortOpts,
+    OverrideEndpointOpts, Peer, PeerContents, RenamePeerOpts, PERSISTENT_KEEPALIVE_INTERVAL_SECS,
 };
 use anyhow::anyhow;
 use colored::*;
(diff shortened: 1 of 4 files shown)


shared/src/prompts.rs
```diff
@@ -396,28 +396,38 @@ pub fn rename_peer(
...
+        let peer_selection: Vec<_> = enabled_peers
+            .iter()
+            .map(|peer| format!("{} ({})", &peer.name, &peer.ip))
+            .collect();
+        let (index, _) = select(
+            &format!("Peer to {}able", if enable { "en" } else { "dis" }),
+            &peer_selection,
+        )?;
+        enabled_peers[index].clone()
+    };
```
strohel, line 421: Super-nit/curiosity question: previously `peer` was `&Peer`, now we've changed it to `Peer` (if I'm looking correctly), needing to `clone()`. Can we keep it a `&Peer`, perhaps if we change `.into_iter()` to `.iter()` in the first arm?

strohel (approved):
Looking good, thanks for your contribution! Added just one supernit/question.

I'd wait a couple of days to that others also have an opportunity to review, but then this is good for merging from me.

shared/src/types.rs
```diff
@@ -348,6 +348,18 @@ pub struct RenamePeerOpts {
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Args)]
+pub struct EnableDisablePeerOpts {
+    /// Name of peer to enable/disable
+    #[clap(long)]
+    pub name: Option<Hostname>,
+
+    /// Bypass confirmation
+    #[clap(long)]
+    pub yes: bool,
```
bschwind, line 360: Am I correct in that the `yes` flag can only really be effectively used alongside the `name` flag? In that case we may want to add a [`requires`](https://docs.rs/clap/4.1.6/clap/builder/struct.Arg.html#method.requires) dependency, though I'm not sure if we can express that with the clap-derive syntax.

bschwind (approved):
Overall this looks good, thanks for the PR! I think this is good to merge after following up on the clarifying questions.

linuskendall:
> Overall this looks good, thanks for the PR! I think this is good to merge after following up on the clarifying questions.

Thanks for the quick review - I have been travelogn but will take a look and address the comments next week.

shared/src/types.rs
goodhoko: Per the [docs](https://docs.rs/clap/latest/clap/_derive/#terminology) it should be doable with a mere

```rust 
    #[arg(long, requires("name"))]
    pub yes: bool,
```

shared/src/prompts.rs
linuskendall: It looks like those clones were just superfluous. Removed.

shared/src/types.rs
linuskendall: Thanks, I made the change accordingly.

linuskendall:
> Overall this looks good, thanks for the PR! I think this is good to merge after following up on the clarifying questions.

Sorry for leaving this dangling for so long! I just lost of track of it. Updated as per the comments.

strohel (approved):
Looking great, thanks, @linuskendall! Good to merge, though would be good to test in real life first.

Want anyone in the community to quickly test? Ideally both the server and client variants of the new options.

linuskendall:
Good to test. For reference - we have been running this patch for the last
6 months or so on our innernet instance!

(Deployed it and kind of forgot that it was a patch we added, so forgot
about this PR! Sorry!)

goodhoko:
@strohel Taking you attention back to this. I think this could me deemed as tested. Can we merge?

strohel:
Thanks for ping, I've hit the merge button. Thanks everybody contributing ans sorry for the delays!

PR #251, opened 2023-03-27
Remove extraneous into_iter() (clippy 1.68 fix)

strohel:
https://doc.rust-lang.org/std/fs/fn.read_dir.html already returns an Iterator (`ReadDir`s IntoIter implementation returns `self`)

It looks like the only PRs I submit are clippy fixes, but I believe one day it will be something more interesting. :)


bschwind (approved):
LGTM!

PR #252, opened 2023-04-14
meta: update jake's email address

mcginty:
Transitioning away from the .su domain email.

Also might recommend that @strohel @bschwind and @skywhale add your names as authors :).


strohel (approved):
Good call, will add us separately

PR #257, opened 2023-05-24
[WIP] Add association commands to server CLI; refactor common commands to shared crate

### LINKED ISSUE #199: add-association from coordination server?
It doesn't seem like it's currently possible to add-association from the coordination innernet-server. Is there a reason why it needs to be done from an admin client?

agausmann:
Resolves #199 

I am refactoring the common CLI commands so that the UI is only implemented once, in the shared crate, and the client and server crates implement a trait to provide the low-level APIs for interface management. This should help improve feature parity in the future, and provides a stronger guarantee for a consistent UI  between the client and server CLIs.

This will bring the `add-association`, `delete-association` and `list-associations` commands to the server CLI (#199).

To Do:

- [x] Implement association commands on server-side
- [x] Migrate tree printing utilities to the shared crate, so the tree mode of `list-cidrs` can still work.
- [ ] Maybe refactor the `show` command to common commands?


strohel:
Thanks @agausmann, this is definitely something we want, and sharing code between client and server to be DRY and to guarantee same UX is very reasonable. I'll review once you mark this as ready.

agausmann:
Update on where I'm at:
I spent a couple days trying to move the `show` command, but it is challenging because it is different from the other commands:

- It can access and display _either_ a single interface or all interfaces.

- On the client side, it uses the local DB instead of the server API, unlike the admin commands.
    (One thing to note: The client-side `list-cidrs` command also used the local DB before the refactor, I'm not sure if that is still desired. My guess is that before, it showed the CIDRs that the peer can access, and now it shows all CIDRs (and may require admin access).

The abstraction I've written doesn't really handle multiple interfaces or the distinction between local/remote data queries.

strohel:
> I spent a couple days trying to move the `show` command, but it is challenging because it is different from the other commands:
> 
>     * It can access and display _either_ a single interface or all interfaces.
> 
>     * On the client side, it uses the local DB instead of the server API, unlike the admin commands.
>       (One thing to note: The client-side `list-cidrs` command also used the local DB before the refactor, I'm not sure if that is still desired. My guess is that before, it showed the CIDRs that the peer can access, and with this current change, it shows all CIDRs (and may require admin access).

Oh I see. Maybe these should be separate commands then?

My feeling is that client commands that use the local cache should preserve this property. Innernet network should continue to operate even if the server is (temporarily) down, and that would ideally include commands that don't need to by design talk to the server.

@mcginty do you have thoughts on this?

mcginty:
Yeah, there probably needs to be better documentation (or just... decisions) to decide when exactly a cache is used or not by a client.

I think @agausmann hit the nail on the head that the main challenge here is to build an abstraction that can access the data via sqlite or HTTP API. There is still a world that could exist where the server accesses its own API, but if I remember right when trying to do that originally it felt very off from a security perspective. Perhaps there's a path forward still with that if the server also listened on a protected unix socket, for example, though.

agausmann:
Sorry this has stalled for so long. I think I'm going to take another stab at this, and re-do it from the latest `main`.

agausmann:
> Maybe these should be separate commands then?

Yes, I think there should be a more clear distinction between "admin" commands and "user" commands. Maybe literally `innernet admin xxx` to make the user's intent clear

PR #258, opened 2023-05-27
Use the proper netlink buffer size with large kernel pages

refi64:
The recommended netlink buffer size is based on the system's page size, which means that the current size is far too small for systems with 16k or 64k pages, such as Asahi Linux or RHEL's kernel-64k for ARM64. On these systems, the server fails to start with errors like this:

Error: Decode error occurred: invalid netlink buffer: length field says 1444 the buffer is 1260 bytes long

Instead, follow the kernel's own netlink docs to compute the buffer size. The approach here matches the approach merged into Chromium recently:

https://chromium-review.googlesource.com/c/chromium/src/+/4312885

### DIFF
--- a/netlink-request/Cargo.toml
+++ b/netlink-request/Cargo.toml
@@ -8,3 +8,5 @@ netlink-sys = "0.8"
 netlink-packet-core = "0.4"
 netlink-packet-generic = "0.3"
 netlink-packet-route = "0.13"
+nix = { version = "0.25", features = ["feature"] }
+once_cell = "1"
--- a/netlink-request/src/lib.rs
+++ b/netlink-request/src/lib.rs
@@ -1,9 +1,5 @@
 #[cfg(target_os = "linux")]
 mod linux {
-    pub const MAX_NETLINK_BUFFER_LENGTH: usize = 4096;
-    pub const MAX_GENL_PAYLOAD_LENGTH: usize =
-        MAX_NETLINK_BUFFER_LENGTH - NETLINK_HEADER_LEN - GENL_HDRLEN;
-
     use netlink_packet_core::{
         NetlinkDeserializable, NetlinkMessage, NetlinkPayload, NetlinkSerializable,
         NETLINK_HEADER_LEN, NLM_F_ACK, NLM_F_CREATE, NLM_F_EXCL, NLM_F_REQUEST,
@@ -15,6 +11,8 @@ mod linux {
     };
     use netlink_packet_route::RtnlMessage;
     use netlink_sys::{constants::NETLINK_GENERIC, protocols::NETLINK_ROUTE, Socket};
+    use nix::unistd::{sysconf, SysconfVar};
+    use once_cell::sync::OnceCell;
     use std::{fmt::Debug, io};
 
     macro_rules! get_nla_value {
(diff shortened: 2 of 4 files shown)


netlink-request/src/lib.rs
```diff
@@ -26,6 +23,19 @@ mod linux {
         };
     }
 
+    pub fn get_max_netlink_buffer_length() -> usize {
+        // https://www.kernel.org/doc/html/v6.2/userspace-api/netlink/intro.html#buffer-sizing
+        // "Netlink expects that the user buffer will be at least 8kB or a page
+        // size of the CPU architecture, whichever is bigger.""
+        const MIN_NELINK_BUFFER_LENGTH: usize = 8 * 1024;
+        let page_size = sysconf(SysconfVar::PAGE_SIZE).unwrap().unwrap() as usize;
```
bschwind, line 31: Do you think we should use a default here instead of double unwrapping?

netlink-request/src/lib.rs
```diff
@@ -26,6 +23,19 @@ mod linux {
         };
     }
 
+    pub fn get_max_netlink_buffer_length() -> usize {
```
bschwind, line 26: Not a strong preference but this could just be called `max_netlink_buffer_length()` and the function below could be `max_genl_payload_length()`

netlink-request/src/lib.rs
```diff
@@ -26,6 +23,19 @@ mod linux {
...
+        // https://www.kernel.org/doc/html/v6.2/userspace-api/netlink/intro.html#buffer-sizing
+        // "Netlink expects that the user buffer will be at least 8kB or a page
+        // size of the CPU architecture, whichever is bigger.""
```
bschwind, line 29: Small nitpick

```suggestion
        // size of the CPU architecture, whichever is bigger."
```

bschwind (approved):
Good catch here! The code looks good, I just have some mostly trivial comments on naming and such.

netlink-request/src/lib.rs
refi64: This shouldn't ever fail unless something has gone horribly wrong, so it's probably safer to just unwrap; I added a comment clarifying this.

netlink-request/src/lib.rs
refi64: Updated, also did some variable renames to match.

bschwind (approved):
LGTM++, I'll wait for at least another review before merging.

netlink-request/Cargo.toml
```diff
@@ -8,3 +8,4 @@ netlink-sys = "0.8"
 netlink-packet-core = "0.4"
 netlink-packet-generic = "0.3"
 netlink-packet-route = "0.13"
+nix = "0.25"
```
strohel, line 11: I see that nix has many features, all of them enabled by default. Looks like we can only enable nix cargo feature called `feature` (:srug:) to save a bit of compilation time?

strohel (changes requested):
Looking great! Just one little suggestion wrt dependencies.

I guess the `sysconf()` call is rather cheap and we don't have to think about caching it (beyond local variable caching already present in this PR)?

mcginty (approved):
Nice comments, nice fix!

+1 for @strohel's suggestion of not using default features in `nix`.

On my machine, `sysconf` is <1ms but is called >200 times given that it's called in a pretty low-level function. I don't think it's a bad idea just from a cleanliness perspective to throw it in a [`OnceCell`](https://github.com/matklad/once_cell), especially since Rust will likely stabilize a similar API in `std` soon enough.

mcginty (approved):
`OnceCell` change looks good!

Issue #233, opened 2022-09-12
Innernet writes /etc/hosts when only the order of mappings in a section has changed, not the mappings themselves


kevenwyld:
Currently /etc/hosts is written very frequently, nearly every minute, because innernet writes the contents of each section as it is received.  Apparently those checkins, or whatever is triggering HostsBuilder to run, is receiving the list of hosts in the order they respond which is not always the same.  This order is carried through to the order mappings appear in the hosts file.

Would it be possible to sort the list of hosts in each section before checking to see if /etc/hosts requires updating?  This would, I think, change the behavior so that updating /etc/hosts only happens when there is actually an ip/hostname update required and make it much less frequent.

Thanks!

strohel:
Hi @kevenwyld, sorting the lines added to hosts file would be a good idea!

I'm looking at the code, it should be a matter of changing `HostsBuilder`'s `HashMap` to `BTreeMap`:
https://github.com/tonarino/innernet/blob/e98c0659efa5cc42564a7273e3baa11ebaed5396/hostsfile/src/lib.rs#L84

However, even when the hostname list is sorder, it seems that there is currently no detection of `hostname file is already the same as we want to write", so we'd need to add that too (I think we should).

Do you want to try contributing such improvement?

kevenwyld:
Thanks @strohel .  I am up for trying to trying to contribute however this would be my first experience with rust.  I will see what I can come up with but it may take me a few days.

kevenwyld:
Hey, just wanted to say that I'm still working on this, sorry for the silence.  @mcginty helped me out a bit getting up to speed.

evaporei:
Hey @kevenwyld ! Since this seems inactive and the change is pretty simple, I decided to open a PR. #259

kevenwyld:
Hey thanks @evaporei!  Life happened and I didn't have time to get it done.

strohel:
Should be resolved by https://github.com/tonarino/innernet/pull/259.

PR #259, opened 2023-05-27
Optimize /etc/hosts writes

evaporei:
The commit messages and issue should provide enough context for the PR :wink: 

Solves https://github.com/tonarino/innernet/issues/233

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -287,9 +287,12 @@ fn update_hosts_file(
             &format!("{}.{}.wg", peer.contents.name, interface),
         );
     }
-    match hosts_builder.write_to(&hosts_path).with_path(hosts_path) {
+    match hosts_builder.write_to(&hosts_path).with_path(&hosts_path) {
         Ok(has_written) if has_written => {
-            log::info!("updated {} with the latest peers.", "/etc/hosts".yellow())
+            log::info!(
+                "updated {} with the latest peers.",
+                hosts_path.to_string_lossy().yellow()
+            )
         },
         Ok(_) => {},
         Err(e) => log::warn!("failed to update hosts ({})", e),
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -280,8 +280,6 @@ fn update_hosts_file(
     hosts_path: PathBuf,
     peers: &[Peer],
 ) -> Result<(), WrappedIoError> {
-    log::info!("updating {} with the latest peers.", "/etc/hosts".yellow());
-
     let mut hosts_builder = HostsBuilder::new(format!("innernet {interface}"));
     for peer in peers {
         hosts_builder.add_hostname(
@@ -289,9 +287,13 @@ fn update_hosts_file(
             &format!("{}.{}.wg", peer.contents.name, interface),
         );
     }
-    if let Err(e) = hosts_builder.write_to(&hosts_path).with_path(hosts_path) {
-        log::warn!("failed to update hosts ({})", e);
-    }
+    match hosts_builder.write_to(&hosts_path).with_path(hosts_path) {
+        Ok(has_written) if has_written => {
+            log::info!("updated {} with the latest peers.", "/etc/hosts".yellow())
+        },
+        Ok(_) => {},
+        Err(e) => log::warn!("failed to update hosts ({})", e),
+    };
 
     Ok(())
 }
--- a/hostsfile/src/lib.rs
+++ b/hostsfile/src/lib.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::HashMap,
+    collections::BTreeMap,
     fmt,
     fs::OpenOptions,
     io::{self, BufRead, BufReader, ErrorKind, Write},
@@ -81,7 +81,7 @@ impl std::error::Error for Error {
 /// ```
 pub struct HostsBuilder {
     tag: String,
-    hostname_map: HashMap<IpAddr, Vec<String>>,
+    hostname_map: BTreeMap<IpAddr, Vec<String>>,
 }
 
 impl HostsBuilder {
@@ -90,7 +90,7 @@ impl HostsBuilder {
     pub fn new<S: Into<String>>(tag: S) -> Self {
         Self {
             tag: tag.into(),
-            hostname_map: HashMap::new(),
+            hostname_map: BTreeMap::new(),
         }
     }
 
--- a/hostsfile/src/lib.rs
+++ b/hostsfile/src/lib.rs
@@ -1,8 +1,8 @@
 use std::{
     collections::BTreeMap,
     fmt,
-    fs::{File, OpenOptions},
-    io::{self, BufRead, BufReader, ErrorKind, Read, Write},
+    fs::OpenOptions,
+    io::{self, BufRead, BufReader, ErrorKind, Write},
     net::IpAddr,
     path::{Path, PathBuf},
     result,
@@ -172,24 +172,6 @@ impl HostsBuilder {
         Ok(hosts_dir.with_file_name(temp_filename))
     }
 
-    fn has_content_changed(hosts_path: &Path, new_contents: &[u8]) -> io::Result<bool> {
-        let hosts_file = File::open(hosts_path)?;
-
-        if hosts_file.metadata()?.len() != new_contents.len() as u64 {
-            return Ok(true);
-        }
-
-        let hosts_buf = BufReader::new(hosts_file);
-
-        for (b1, b2) in hosts_buf.bytes().zip(new_contents) {
-            if b1.unwrap() != *b2 {
-                return Ok(true);
-            }
-        }
-
-        Ok(false)
-    }
-
     /// Inserts a new section to the specified hosts file.  If there is a section with the same tag
     /// name already, it will be replaced with the new list instead.
     ///
@@ -227,9 +209,33 @@ impl HostsBuilder {
         let begin = lines.iter().position(|line| line.trim() == begin_marker);
         let end = lines.iter().position(|line| line.trim() == end_marker);
 
+        let mut lines_to_insert = vec![];
+        if !self.hostname_map.is_empty() {
+            lines_to_insert.push(begin_marker);
+            for (ip, hostnames) in &self.hostname_map {
+                if cfg!(windows) {
+                    // windows only allows one hostname per line
+                    for hostname in hostnames {
+                        lines_to_insert.push(format!("{ip} {hostname}"));
+                    }
+                } else {
+                    // assume the same format as Unix
+                    lines_to_insert.push(format!("{} {}", ip, hostnames.join(" ")));
+                }
+            }
+            lines_to_insert.push(end_marker);
+        }
+
         let insert = match (begin, end) {
             (Some(begin), Some(end)) => {
-                lines.drain(begin..end + 1);
+                let old_section: Vec<String> = lines.drain(begin..end + 1).collect();
+
+                let sections_are_eq = old_section == lines_to_insert;
+
+                if sections_are_eq {
+                    return Ok(false);
+                }
+
                 begin
             },
             (None, None) => {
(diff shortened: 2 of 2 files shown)


bschwind (approved):
Overall this looks good, thanks for picking it up!

alerque:
This will be a welcome change reducing the churn in system diff files and incremental backups. I have a few pretty "still" server setups where little or nothing changes in etc, but innernet has been consistently shuffling the cheese in hosts files causing unnecessary system change tracking.

bschwind:
There's also a separate issue I want to look into, where repeated calls to `innernet fetch` will result in the server claiming a bunch of peers have been modified, even though it's very unlikely they actually changed. Unrelated to this PR though.

hostsfile/src/lib.rs
```diff
@@ -171,6 +171,24 @@ impl HostsBuilder {
...
+            return Ok(true);
+        }
+
+        let hosts_buf = BufReader::new(hosts_file);
+
+        for (b1, b2) in hosts_buf.bytes().zip(new_contents) {
+            if b1.unwrap() != *b2 {
+                return Ok(true);
+            }
+        }
```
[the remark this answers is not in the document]
evaporei: Thanks for the suggestion! It also requires a `!` before it though, since the logic is inversed.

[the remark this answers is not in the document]
bschwind: Oops, yep you're right. I'll let you decide which method you'd prefer. The nice thing about `Iterator::eq` is it will include the length check, though I understand why you're doing the length check ahead of time so you don't have to read the file.

The length check before reading could _theoretically_ be racy, but nothing to worry about. To trigger it, I think a new `/etc/hosts` would have to be written which has a truncated version of `new_contents` in between the metadata length check and the byte comparison. The end result would be that you just don't write the file at all, which is no big deal I suppose.

hostsfile/src/lib.rs
```diff
@@ -265,10 +283,12 @@ impl HostsBuilder {
     }
 
     fn write_and_swap(temp_path: &Path, hosts_path: &Path, contents: &[u8]) -> io::Result<()> {
-        // Copy the file we plan on modifying so its permissions and metadata are preserved.
-        std::fs::copy(hosts_path, temp_path)?;
-        Self::write_clobber(temp_path, contents)?;
-        std::fs::rename(temp_path, hosts_path)?;
+        if Self::has_content_changed(hosts_path, contents)? {
```
bschwind, line 286: Just a thought but not necessarily something we should do in this PR:

Have these write functions return a `bool` indicating if we actually wrote the hosts file, and only output a message if we do.

Currently `innernet` unconditionally prints:

```
[*] updating /etc/hosts with the latest peers.
```

Even if there is nothing to change.

evaporei: That's a great idea!

So I took a look at `client/src/main.rs` and thought about moving the `updating /etc/hosts` log to an `else` statement [here](https://github.com/tonarino/innernet/blob/ae96e05e903a9946f9da4579baac11fef7ad9ecf/client/src/main.rs#L294), also the message should say `updated /etc/hosts` instead.

Do you think there should be a log at the beginning of `update_hosts_file` with something like `trying to update /etc/hosts` or should we use this moment to reduce the number of logs related to `/etc/hosts`?

bschwind: Just my opinion, but I think we should only output something related to `/etc/hosts` if we actually do something to it.

Right now we output `updating /etc/hosts with the latest peers`, but it's not like the user has any chance to stop it before it writes, so we might as well just output a message when we update it, or not output anything if we don't.

@mcginty @strohel what do you think?

strohel (commented):
Thanks for picking this up! This would be a nice addition.

Functionality-wise the current version looks good (modulo one stale debug log + a suggestion to tweak the other loogging). But I've added a comment that suggest doing the comparison at a slightly different level, curious what you'd think.

hostsfile/src/lib.rs
evaporei: Fixed!

hostsfile/src/lib.rs
```diff
@@ -171,6 +171,24 @@ impl HostsBuilder {
...
     }
 
+    fn has_content_changed(hosts_path: &Path, new_contents: &[u8]) -> io::Result<bool> {
+        let hosts_file = File::open(hosts_path)?;
+
+        if hosts_file.metadata()?.len() != new_contents.len() as u64 {
+            return Ok(true);
+        }
+
+        let hosts_buf = BufReader::new(hosts_file);
```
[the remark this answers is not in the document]
evaporei: Indeed the file doesn't have to be read again! I just implemented your suggestion in a new commit, see what you think of it :)

hostsfile/src/lib.rs
```diff
@@ -206,9 +210,39 @@ impl HostsBuilder {
...
+            return Ok(false);
+        }
+
         let insert = match (begin, end) {
             (Some(begin), Some(end)) => {
-                lines.drain(begin..end + 1);
+                // Add '\n's in between lines
+                let old_section: String = lines
+                    .drain(begin..end + 1)
+                    .fold(String::new(), |acc, line| acc + &line + "\n");
```
evaporei, line 238: I would like to use `intersperse`, but that's only available on the nightly channel :/

hostsfile/src/lib.rs
```diff
@@ -206,9 +210,39 @@ impl HostsBuilder {
...
+
+                let sections_are_eq = old_section == from_utf8(&lines_to_insert).unwrap();
```
evaporei, line 240: @strohel @bschwind Feel free to suggest another way of comparing this!

hostsfile/src/lib.rs
```diff
@@ -206,9 +210,39 @@ impl HostsBuilder {
...
+                } else {
+                    // assume the same format as Unix
+                    writeln!(lines_to_insert, "{} {}", ip, hostnames.join(" "))?;
+                }
+            }
+            writeln!(lines_to_insert, "{end_marker}")?;
+        } else {
+            // There's no need to write to the hosts file if the map is empty
+            return Ok(false);
+        }
```
strohel, line 231: I think there is one corner case: when the map _becomes_ empty. Maybe this cannot happen in innernet as the current peer is probably always online, but `hostsfile` is a semi-independent crate, so it would be better not to have this assumption.

The old code still wrote the file but skipped the begin and end markers, which is I think a good behavior.

hostsfile/src/lib.rs
```diff
@@ -206,9 +210,39 @@ impl HostsBuilder {
         let begin = lines.iter().position(|line| line.trim() == begin_marker);
         let end = lines.iter().position(|line| line.trim() == end_marker);
 
+        let mut lines_to_insert = vec![];
+        if !self.hostname_map.is_empty() {
+            writeln!(lines_to_insert, "{begin_marker}")?;
```
strohel, line 215: Currently `lines_to_insert` is `Vec<u8>`, but I think it would simplify code below if it were a `Vec<String>`. I.e. we would be pushing lines as Strings to the Vec as opposed to writing it to a continuous buffer.

It should make the comparison with `lines` below more straightforward.

client/src/main.rs
```diff
@@ -280,8 +280,6 @@ fn update_hosts_file(
     hosts_path: PathBuf,
     peers: &[Peer],
 ) -> Result<(), WrappedIoError> {
-    log::info!("updating {} with the latest peers.", "/etc/hosts".yellow());
```
strohel, line 283: BTW I think the intention of @bschwind was to move (and reword) this info log, not to completely remove it?

hostsfile/src/lib.rs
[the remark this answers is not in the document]
strohel: Yeah looking better, though I've added a comment that I think that using `lines_to_insert: Vec<String>` (per above) instead of current `lines_to_insert: Vec<u8>` could simplify things even more.

strohel (changes requested):

hostsfile/src/lib.rs
mcginty: In other parts of the code, we've put something like:

```rust
// TODO: use `intersperse()` when stable, tracking issue: https://github.com/rust-lang/rust/issues/XXX
...
```

So that we can periodically see if new language features allow us to clean up the code eventually :).

client/src/main.rs
bschwind: Right, we should print out a message if we actually wrote new data to the hosts file.

hostsfile/src/lib.rs
bschwind: This definitely seems like the best suggestion for simplifying the code here.

client/src/main.rs
evaporei: Fixed!

hostsfile/src/lib.rs
evaporei: `else` removed :heavy_check_mark:

hostsfile/src/lib.rs
evaporei: Since `intersperse` won't be needed anymore, I won't add the comment, but it's a good habit to add those :ok_hand:

client/src/main.rs
```diff
@@ -280,18 +280,20 @@ fn update_hosts_file(
...
             peer.contents.ip,
             &format!("{}.{}.wg", peer.contents.name, interface),
         );
     }
-    if let Err(e) = hosts_builder.write_to(&hosts_path).with_path(hosts_path) {
-        log::warn!("failed to update hosts ({})", e);
-    }
+    match hosts_builder.write_to(&hosts_path).with_path(hosts_path) {
+        Ok(has_written) if has_written => {
+            log::info!("updated {} with the latest peers.", "/etc/hosts".yellow())
```
bschwind, line 292: This will cause a borrow error on the line above for `with_path(hosts_path)`, so that should also be changed to `.with_path(&hosts_path)`

```suggestion
            log::info!("updated {} with the latest peers.", hosts_path.to_string_lossy().yellow())
```

bschwind (commented):
This is looking really nice and clean, good work!

I found one more thing to change, and then I'll leave the final reviews to @mcginty and @strohel

client/src/main.rs
bschwind: I realized this was a bug in the original code!

evaporei: Done!

bschwind (approved):
Final LGTM from me, I'll pass it along to @mcginty and @strohel

client/src/main.rs
```diff
@@ -280,18 +280,23 @@ fn update_hosts_file(
...
         hosts_builder.add_hostname(
             peer.contents.ip,
             &format!("{}.{}.wg", peer.contents.name, interface),
         );
     }
-    if let Err(e) = hosts_builder.write_to(&hosts_path).with_path(hosts_path) {
-        log::warn!("failed to update hosts ({})", e);
-    }
+    match hosts_builder.write_to(&hosts_path).with_path(&hosts_path) {
+        Ok(has_written) if has_written => {
```
strohel, line 291: Nice touch, this is much clearer/more explicit than an alternative to match direclty `Ok(true)`.

hostsfile/src/lib.rs
```diff
@@ -206,9 +209,33 @@ impl HostsBuilder {
...
+        }
+
         let insert = match (begin, end) {
             (Some(begin), Some(end)) => {
-                lines.drain(begin..end + 1);
+                let old_section: Vec<String> = lines.drain(begin..end + 1).collect();
+
+                let sections_are_eq = old_section == lines_to_insert;
+
+                if sections_are_eq {
```
strohel, line 235: This should be now clear enough not to introduce and intermediate variable:
```suggestion
                if old_section == lines_to_insert {
```

strohel (approved):
Perfect! Added one last nit, but approved already. Thanks for all the iterations.

Before merging I'll give this some testing in the coming days.

evaporei:
Thank you all for the thoughtful reviews and suggestions :blush:

strohel:
Just quickly tested, works well for me, merging.

PR #262, opened 2023-05-28
Reset peer's endpoint when NAT traversal fails to connect to any endpoint candidates

### LINKED ISSUE #260: Peers are reported as "modified" when fetching repeatedly
On one of my networks, if I issue fetch commands to the server one after the other, I will get the same set of peers reported as "modified".

Example output (with names changed to generic names):

```
[brian@machine ~]$ sudo innernet fetch NETWORK_NAME
[sudo] password for brian: 
[*] fetching state for NETWORK_NAME from server...
[*]   peer peer_1 (KEY_1...) was modified.
[*]   peer peer_2 (KEY_2...) was modified.
[*]   peer peer_3 (KEY_3...) was modified.
[*]   peer peer_4 (KEY_4...) was modified.
[*]   peer peer_5 (KEY_5...) was modified.
[*]   peer peer_6 (KEY_6...) was modified.
[*]   peer peer_7 (KEY_7...) was modified.
[*] updating /etc/hosts with the latest peers.

[*] updated interface NETWORK_NAME

[*] reporting 2 interface addresses as NAT traversal candidates
[*] Attempting to establish connection with 7 remaining unconnected peers...
[brian@machine ~]$ sudo innernet fetch NETWORK_NAME
[*] fetching state for NETWORK_NAME from server...
[*]   peer peer_1 (KEY_1...) was modified.
[*]   peer peer_2 (KEY_2...) was modified.
[*]   peer peer_3 (KEY_3...) was modified.
[*]   peer peer_4 (KEY_4...) was modified.
[*]   peer peer_5 (KEY_5...) was modified.
[*]   peer peer_6 (KEY_6...) was modified.
[*]   peer peer_7 (KEY_7...) was modified.
[*] updating /etc/hosts with the latest peers.

[*] updated interface NETWORK_NAME

[*] reporting 2 interface addresses as NAT traversal candidates
[*] Attempting to establish connection with 7 remaining unconnected peers...
[brian@machine ~]$ sudo innernet fetch NETWORK_NAME
[*] fetching state for NETWORK_NAME from server...
[*]   peer peer_1 (KEY_1...) was modified.
[*]   peer peer_2 (KEY_2...) was modified.
[*]   peer peer_3 (KEY_3...) was modified.
[*]   peer peer_4 (KEY_4...) was modified.
[*]   peer peer_5 (KEY_5...) was modified.
[*]   peer peer_6 (KEY_6...) was modified.
[*]   peer peer_7 (KEY_7...) was modified.
[*] updating /etc/hosts with the latest peers.

[*] updated interface NETWORK_NAME

[*] reporting 2 interface addresses as NAT traversal candidates
[*] Attempting to establish connection with 7 remaining unconnected peers...
```

These 3 fetches were issued one after the other, so it is very unlikely the exact same set of peers happened to change in between fetches.

bschwind:
Closes #260

We had the logic already to reset a peer's endpoint to the server-provided endpoint when no endpoint candidates were viable, but we were missing a call to it at the end of `nat.rs`.

Otherwise the client was always setting the endpoint to the last endpoint in the list of NAT candidates given from the server, which triggered a diff on every fetch.

I think this was just an oversight from the original commit here:

https://github.com/tonarino/innernet/pull/134/commits/9e55d806e412c083a5905210f8e61cd22a7d7d9e ([PR](https://github.com/tonarino/innernet/pull/134/files#diff-7c27d7c09358d86ce07c74518a89167534922c300a0e2f2829fceb2f05128c39R107))

### DIFF
--- a/client/src/nat.rs
+++ b/client/src/nat.rs
@@ -43,7 +43,11 @@ impl<'a> NatTraverse<'a> {
             backend,
             remaining,
         };
-        nat_traverse.refresh_remaining()?;
+
+        // We don't need to reset peer endpoints here as we have not changed their
+        // endpoint to a candidate yet.
+        let _exhausted = nat_traverse.refresh_remaining()?;
+
         Ok(nat_traverse)
     }
 
@@ -91,9 +95,7 @@ impl<'a> NatTraverse<'a> {
         let exhausted = self.refresh_remaining()?;
 
         // Reset peer endpoints that had no viable candidates back to the server-reported one, if it exists.
-        let reset_updates = exhausted
-            .into_iter()
-            .filter_map(|peer| set_endpoint(&peer.public_key, peer.endpoint.as_ref()));
+        let reset_updates = peer_reset_updates(exhausted);
 
         // Set all peers' endpoints to their next available candidate.
         let candidate_updates = self.remaining.iter_mut().filter_map(|peer| {
@@ -110,17 +112,35 @@ impl<'a> NatTraverse<'a> {
             .add_peers(&updates)
             .apply(self.interface, self.backend)?;
 
+        let mut peers_to_reset = vec![];
         let start = Instant::now();
         while start.elapsed() < STEP_INTERVAL {
-            self.refresh_remaining()?;
+            let mut exhausted = self.refresh_remaining()?;
+            peers_to_reset.append(&mut exhausted);
+
             if self.is_finished() {
                 log::debug!("NAT traverser is finished!");
                 break;
             }
             std::thread::sleep(Duration::from_millis(100));
         }
+
+        // For any peers which didn't connect and have no more candidates
+        // to try, reset their endpoint to the server-provided endpoint.
+        let reset_updates: Vec<_> = peer_reset_updates(peers_to_reset).collect();
+
+        DeviceUpdate::new()
+            .add_peers(&reset_updates)
+            .apply(self.interface, self.backend)?;
+
         Ok(())
     }
+}
+
+fn peer_reset_updates(peers_to_reset: Vec<Peer>) -> impl Iterator<Item = PeerConfigBuilder> {
+    peers_to_reset
+        .into_iter()
+        .filter_map(|peer| set_endpoint(&peer.public_key, peer.endpoint.as_ref()))
 }
 
 /// Return a PeerConfigBuilder if an endpoint exists and resolves successfully.
--- a/client/src/nat.rs
+++ b/client/src/nat.rs
@@ -33,10 +33,19 @@ impl<'a> NatTraverse<'a> {
             // Limit reported alternative candidates to 10.
             peer.candidates.truncate(10);
 
-            // remove server-reported endpoint from elsewhere in the list if it existed.
+            // Remove server-reported endpoint from elsewhere in the list if it existed.
             let endpoint = peer.endpoint.clone();
             peer.candidates
                 .retain(|addr| Some(addr) != endpoint.as_ref());
+
+            // Add the server-reported endpoint to the beginning of the list. In the event
+            // no other endpoints worked, the remaining endpoint in the list will be the one
+            // assigned to the peer so it should default to the server-reported endpoint.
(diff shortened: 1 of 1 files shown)


client/src/nat.rs
```diff
@@ -43,7 +43,11 @@ impl<'a> NatTraverse<'a> {
             backend,
             remaining,
         };
-        nat_traverse.refresh_remaining()?;
+
+        // We don't need to reset peer endpoints here as we have not changed their
+        // endpoint to a candidate yet.
+        let _exhausted = nat_traverse.refresh_remaining()?;
```
strohel, line 49: Would it be too hacky to add `peer.endpoint` at the end of `peer.candidates` for each peer with non-empty candidates here?

It has a potential of simiplifiying logic (no need to collect `peers_to_reset`) in this PR but also in the preexisting code: I think that `refresh_remaining()` for example wouldn't have to return `exhaused`.

client/src/nat.rs
```diff
@@ -110,19 +112,37 @@ impl<'a> NatTraverse<'a> {
...
+
+        DeviceUpdate::new()
+            .add_peers(&reset_updates)
+            .apply(self.interface, self.backend)?;
+
         Ok(())
     }
 }
 
+fn peer_reset_updates(peers_to_reset: Vec<Peer>) -> impl Iterator<Item = PeerConfigBuilder> {
```
strohel, line 140: (if this function stays): Perhaps this could be called `reset_endpoints()`?

`peer_reset_updates()` sounds like "reset updates of a peer", but this operates on multiple peers, and sets them to server-reported endpoint. (it is somewhat ambiguous what "update" means here).

I see the name is complicated by the fact this does not actually do the change, rather provides the iterator of commands to do that.

strohel (commented):
Oh yay, I've long wished to do a change along these lines!

One suggestion to slightly different approach to the implementation, though not sure what you'd think of it.

client/src/nat.rs
```diff
@@ -91,9 +95,7 @@ impl<'a> NatTraverse<'a> {
         let exhausted = self.refresh_remaining()?;
 
         // Reset peer endpoints that had no viable candidates back to the server-reported one, if it exists.
-        let reset_updates = exhausted
-            .into_iter()
-            .filter_map(|peer| set_endpoint(&peer.public_key, peer.endpoint.as_ref()));
+        let reset_updates = peer_reset_updates(exhausted);
```
mcginty, line 98: Do we actually need to do this as a device update here when we're resetting them all at the end? I think it might feel cleaner if we just keep track of who needs to be reset and then do it all once the `NatTraverse` is totally done.

bschwind: Ah, do you mean we should collect them in a Vec for now and then reset them all at the end? That would work too.

I think it was originally done here because there were already other device updates happening at this point.

client/src/nat.rs
bschwind: That could work, though the behavior is a bit more implicit. I'll take a look and see how much more straightforward it makes things.

I realized too that people can and will Ctrl-C in the middle of NAT traversal attempts, which could leave peers in a potentially bad state. Should we try to mitigate this, or just expect that the user will `fetch` again if they're having connectivity troubles?

strohel: > I realized too that people can and will Ctrl-C in the middle of NAT traversal attempts, which could leave peers in a potentially bad state. Should we try to mitigate this, or just expect that the user will fetch again if they're having connectivity troubles?

That would be eventually nice to handle, but maybe postpone for a follow-up PR in order not to complicate this one? Unless you're already into it.

bschwind: @strohel I tried out your suggestion and I like the big simplifications it provided so I pushed a commit with that change.

strohel (approved):
Oh nice! Not counting the comment (which is nice and needed), this even resulted in net code (logic) removal.

client/src/nat.rs
strohel: Yay!

mcginty (approved):
Logic looks cleaner, I like it!

PR #263, opened 2023-05-28
Turn ChangeString into a PeerChange enum

### LINKED ISSUE #261: Replace ChangeString with an enum
The [ChangeString](https://github.com/tonarino/innernet/blob/ae96e05e903a9946f9da4579baac11fef7ad9ecf/shared/src/types.rs#L523-L527) type is mainly just used for presentation to the user in `verbose` mode to explain which properties of a peer changed.

It currently uses Strings, but since there are a finite number of properties that could change, and we know what they all are, might as well use an enum.

I'd like them to be enums as well because I want to detect certain changes and suppress/modify the output.

If a peer is not reachable but the server has reported endpoint candidates, `innernet fetch` will claim the peer is "modified".

I would prefer to not print "modified" but instead something along the lines of "attempting to connect to peer X". And this should only be printed if NAT traversal is enabled.

Alternatively, if the actual list of endpoint candidates has changed then it would be good to explicitly mention that. Right now the condition is `!endpoint_changed && !new.candidates.is_empty()`.

bschwind:
...So that we don't print NAT traversal reattempts as a "modification" on the peer. This should reduce "Peer X was modified" messages which can occur over and over on subsequent calls to `innernet fetch`

Closes #261

### DIFF
--- a/client/src/util.rs
+++ b/client/src/util.rs
@@ -3,7 +3,9 @@ use colored::*;
 use indoc::eprintdoc;
 use log::{Level, LevelFilter};
 use serde::{de::DeserializeOwned, Serialize};
-use shared::{interface_config::ServerInfo, Interface, PeerDiff, INNERNET_PUBKEY_HEADER};
+use shared::{
+    interface_config::ServerInfo, Interface, PeerChange, PeerDiff, INNERNET_PUBKEY_HEADER,
+};
 use std::{ffi::OsStr, io, path::Path, time::Duration};
 use ureq::{Agent, AgentBuilder};
 
@@ -137,13 +139,30 @@ pub fn permissions_helptext(config_dir: &Path, data_dir: &Path, e: &io::Error) {
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq)]
+enum ChangeAction {
+    Added,
+    Modified,
+    Removed,
+}
+
+impl ChangeAction {
+    fn colored_output(&self) -> ColoredString {
+        match self {
+            Self::Added => "added".green(),
+            Self::Modified => "modified".green(),
+            Self::Removed => "removed".green(),
+        }
+    }
+}
+
 pub fn print_peer_diff(store: &DataStore, diff: &PeerDiff) {
     let public_key = diff.public_key().to_base64();
 
-    let text = match (diff.old, diff.new) {
-        (None, Some(_)) => "added".green(),
-        (Some(_), Some(_)) => "modified".yellow(),
-        (Some(_), None) => "removed".red(),
+    let change_action = match (diff.old, diff.new) {
+        (None, Some(_)) => ChangeAction::Added,
+        (Some(_), Some(_)) => ChangeAction::Modified,
+        (Some(_), None) => ChangeAction::Removed,
         _ => unreachable!("PeerDiff can't be None -> None"),
     };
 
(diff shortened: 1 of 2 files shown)


client/src/util.rs
```diff
@@ -158,15 +177,30 @@ pub fn print_peer_diff(store: &DataStore, diff: &PeerDiff) {
...
+    if change_action == ChangeAction::Modified
+        && diff
+            .changes()
+            .iter()
+            .all(|c| *c == PeerChange::NatTraverseReattempt)
+    {
+        // If this peer was "modified" but the only change is a NAT Traversal Reattempt,
+        // don't bother printing this peer.
+        return;
+    }
```
bschwind, line 189: This was what resulted in lots of "peer was modified" messages, because the `NatTraverseReattempt` "change" was somewhat of a hack to trigger NAT traversal attempts for that peer.

It might actually make more sense to remove this `PeerChange::NatTraverseReattempt` "change" (which again, isn't actually a change) and instead just collect all unconnected peers that have endpoint candidates and try to connect to them after calls to fetch.

That would probably remove a decent amount of special-cased code.

client/src/util.rs
```diff
@@ -158,15 +177,30 @@ pub fn print_peer_diff(store: &DataStore, diff: &PeerDiff) {
...
         "  peer {} ({}...) was {}.",
         peer_name.yellow(),
         &public_key[..10].dimmed(),
-        text
+        change_action.colored_output(),
     );
 
     for change in diff.changes() {
-        log::debug!("    {}", change);
+        if let PeerChange::Endpoint { .. } = change {
```
bschwind, line 199: I felt that endpoint changes merited an info print instead of debug. When I see that a peer was modified I typically want to know if they got a sensible endpoint.

shared/src/types.rs
```diff
@@ -526,6 +526,59 @@ pub struct ChangeString {
     new: Option<String>,
 }
```
strohel, line 527: Is ChangeString still used?

strohel (commented):
@bschwind just to understand this better before I fully review: did #262 remove all the spurious "peer modified" messages, or not yet and this fix is needed to make the logs "completely nice"?

Turning `ChangeString` into less dynamic enum makes good sense to me, I just need to think more about the functional changes.

bschwind:
> did https://github.com/tonarino/innernet/pull/262 remove all the spurious "peer modified" messages, or not yet and this fix is needed to make the logs "completely nice"?

No, there were still spurious "peer modified" messages because it counts a "NAT traversal reattempt" as a modification of the peer.

mcginty:
Nice, so much cleaner. I didn't look deeper than this diff but I definitely agree if "NAT Traverse Reattempt" is all that's adding these extra lines and not actually producing any useful information to the user, we should just get rid of it.

shared/src/types.rs
bschwind: Ah thank you, I intended to remove this and forgot!

bschwind:
Rebased on the latest `main`

mcginty (approved):

strohel (commented):
Porst-merge LGTM! Sorry I was too slow and didn't manage to submit my pending comments before merge, but good you went ahead.

It wouldn't be me if I didn't add one optional supernit.

PR #264, opened 2023-05-30
migrate from lazy_static to once_cell across project

mcginty:
Carrying over a suggestion from #258 to the whole codebase - it looks like `std` will eventually integrate something like this crate, and might as well not have two of basically the same dependency in the mean time.

(Also got rid of a compiler warning about an unneeded `mut`.)


client/Cargo.toml
```diff
@@ -32,6 +31,7 @@ ureq = { version = "2", default-features = false, features = ["json"] }
 wireguard-control = { path = "../wireguard-control" }
 
 [dev-dependencies]
+lazy_static = "1"
```
strohel, line 34: There's still one last usage of `lazy_static!` in `data_store.rs` tests - is that intentional?

strohel (approved):
Thanks, this is much more ergonomic. Just one last question.

client/Cargo.toml
mcginty: Ok, updated :).

strohel: Hehe, you're quick, thanks. .)

PR #265, opened 2023-05-31
Various dependency updates

mcginty:
This was a long time coming, and I figured I might as well be the one to take care of some of the boring stuff :).

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -1,5 +1,5 @@
 use anyhow::{anyhow, bail};
-use clap::{AppSettings, Args, IntoApp, Parser, Subcommand};
+use clap::{Args, Parser, Subcommand, ArgAction};
 use colored::*;
 use dialoguer::{Confirm, Input};
 use hostsfile::HostsBuilder;
@@ -47,15 +47,14 @@ macro_rules! println_pad {
 }
 
 #[derive(Clone, Debug, Parser)]
-#[clap(name = "innernet", author, version, about)]
-#[clap(global_setting(AppSettings::DeriveDisplayOrder))]
+#[command(name = "innernet", author, version, about)]
 struct Opts {
     #[clap(subcommand)]
     command: Option<Command>,
 
     /// Verbose output, use -vv for even higher verbositude
-    #[clap(short, long, parse(from_occurrences))]
-    verbose: u64,
+    #[clap(short, long, action = ArgAction::Count)]
+    verbose: u8,
 
     #[clap(short, long, default_value = "/etc/innernet")]
     config_dir: PathBuf,
@@ -254,7 +253,7 @@ enum Command {
 
     /// Generate shell completion scripts
     Completions {
-        #[clap(arg_enum)]
+        #[clap(value_enum)]
         shell: clap_complete::Shell,
     },
 }
@@ -1170,7 +1169,7 @@ fn print_peer(peer: &PeerState, short: bool, level: usize) {
 
 fn main() {
     let opts = Opts::parse();
-    util::init_logger(opts.verbose);
+    util::init_logger(opts.verbose as u64);
 
     let argv0 = std::env::args().next().unwrap();
     let executable = Path::new(&argv0).file_name().unwrap().to_str().unwrap();
@@ -1275,6 +1274,7 @@ fn run(opts: &Opts) -> Result<(), Error> {
             override_endpoint(&interface, opts, sub_opts)?;
         },
         Command::Completions { shell } => {
+            use clap::CommandFactory;
             let mut app = Opts::command();
             let app_name = app.get_name().to_string();
             clap_complete::generate(shell, &mut app, app_name, &mut std::io::stdout());
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -73,7 +73,7 @@ struct HostsOpt {
     hosts_path: PathBuf,
 
     /// Don't write to any hosts files
-    #[clap(long = "no-write-hosts", conflicts_with = "hosts-path")]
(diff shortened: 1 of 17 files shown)


netlink-request/Cargo.toml
```diff
@@ -4,9 +4,10 @@ version = "1.5.5"
...
 [target.'cfg(target_os = "linux")'.dependencies]
-netlink-sys = "0.8"
-netlink-packet-core = "0.4"
-netlink-packet-generic = "0.3"
-netlink-packet-route = "0.13"
+netlink-sys = "0.8.5"
+netlink-packet-core = "0.5"
+netlink-packet-generic = "0.3.2"
+netlink-packet-route = "0.15.0"
+netlink-packet-utils = "0.5.2"
```
bschwind, line 11: Was there something released which requires us to specify the patch versions?

shared/src/netlink.rs
```diff
@@ -23,14 +26,12 @@ fn if_nametoindex(interface: &InterfaceName) -> Result<u32, io::Error> {
...
 pub fn set_up(interface: &InterfaceName, mtu: u32) -> Result<(), io::Error> {
     let index = if_nametoindex(interface)?;
-    let message = LinkMessage {
-        header: LinkHeader {
-            index,
-            flags: IFF_UP,
-            ..Default::default()
-        },
-        nlas: vec![link::nlas::Nla::Mtu(mtu)],
-    };
```
bschwind, line 33: What was wrong with this original code?

wireguard-control/Cargo.toml
```diff
@@ -10,17 +10,18 @@ repository = "https://github.com/tonarino/innernet"
...
 [dependencies]
-base64 = "0.13"
-hex = "0.4"
+base64 = "0.13.1"
+hex = "0.4.3"
 libc = "0.2"
 log = "0.4"
 rand_core = { version = "0.6", features = ["getrandom"] }
-curve25519-dalek = "4.0.0-pre.2"
+x25519-dalek = { version = "=2.0.0-rc.2", features = ["static_secrets"] }
```
bschwind, line 18: Guessing this is a better alternative to `curve25519-dalek`?

netlink-request/Cargo.toml
strohel: Yeah this reminds me a discussion with @goodhoko in https://github.com/tonarino/acoustic_profiler/pull/12#discussion_r1172379880.

As innernet is shipped as a binary and not a library, I don't think we need to aim for maximum possible version range for dependencies (which is tedious if done correctly, per reasoning in the linked thread). So I'm cool with specifying also the patch versions. It reduces the combinatorial space of possible build configurations.

Only thing that tickles my OCD is saying `0.15.0` instead of `0.15` (these 2 are strictly equivalent), but OTOH that is consistent with other lines..

shared/src/netlink.rs
strohel: FWIW, I for one would have also preferred the previous style. It allows not having the variables `mut`, i.e. it encodes tiny bit more information in the source code.

Extracting nested structs to own variables is very fine, so the following would be also cool (but not advocating for it):
```rust
    let header = LinkHeader {
        index,
        flags: IFF_UP,
        ..Default::default()
    };
    let message = LinkMessage {
        header,
        nlas: vec![link::nlas::Nla::Mtu(mtu)],
    };
```

strohel (commented):
This was long overdue, thanks a lot! Just adding some questions and follow-ups to Brian's comments.

netlink-request/Cargo.toml
mcginty: I've gotten into the habit after discussions at work about this - specifying a higher and more specific semver prevents any bugs in previous versions from showing up when cargo decides for whatever reason to use it instead. For example, I actually had a bug with an earlier patch release in the netlink crate family not compiling.

Happy to get rid of that extra '0' for you @strohel :).

shared/src/netlink.rs
mcginty: I would also have preferred the previous style but they got rid of it, since they labeled all structs "non-exhuastive" to make their compatibility story simpler for them. Not my favorite tbh.

client/src/main.rs
```diff
@@ -1170,7 +1169,7 @@ fn print_peer(peer: &PeerState, short: bool, level: usize) {
 
 fn main() {
     let opts = Opts::parse();
-    util::init_logger(opts.verbose);
+    util::init_logger(opts.verbose as u64);
```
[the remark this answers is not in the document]
mcginty: Doy - left that in while I was just playing around with clap changes. Changed.

wireguard-control/src/backends/kernel.rs
```diff
@@ -3,7 +3,7 @@ use crate::{
     PeerConfigBuilder, PeerInfo, PeerStats,
 };
 use netlink_packet_core::{
-    NetlinkMessage, NetlinkPayload, NLM_F_ACK, NLM_F_CREATE, NLM_F_EXCL, NLM_F_REQUEST,
+    NetlinkMessage, NetlinkPayload, NLM_F_ACK, NLM_F_CREATE, NLM_F_DUMP, NLM_F_EXCL, NLM_F_REQUEST,
```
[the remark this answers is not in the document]
mcginty: yep

shared/src/netlink.rs
bschwind: Ahhh, non-exhaustive structs, I see.

bschwind (approved):
LGTM again, thanks!

strohel (commented):
Post-merge LGTM++, thanks!

PR #266, opened 2023-06-02
Drop warning about using the inn symlink

strohel:
Follow-up to just-merged #175. We no longer install the `inn` symlink, so users doing that manually should know what they are doing.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -1176,16 +1176,6 @@ fn main() {
     let opts = Opts::parse();
     util::init_logger(opts.verbose);
 
-    let argv0 = std::env::args().next().unwrap();
-    let executable = Path::new(&argv0).file_name().unwrap().to_str().unwrap();
-    if executable == "inn" {
-        log::warn!("");
-        log::warn!("  {}: the {} shortcut will be removed from OS packages soon in favor of users creating a shell alias.", "WARNING".bold(), "inn".yellow());
-        log::warn!("");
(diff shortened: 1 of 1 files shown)


mcginty (approved):

alerque:
I would actually have held off on this because while end users that make their own links probably know what they are up to, but distro packagers (I'm deep into that world!) are sometimes a little less attentive and I think there are still distro packages that will add this link out of the box and might go on doing so for a while until somebody brings it to their attention.

In any case this PR is obsolete since it was tacked onto #175, and I would treat it like water under the bridge. If any action is taken the best thing to do is probably run through the packaging that is out there (Repology is good for finding them) and directly submit bug reports to any that haven't removed the link.

alerque:
I read the other PR wrong, it was a mention not a commit pushed to the branch. Disregard.

As for timing you might be right, I forgot this has been in several releases already along with the documentation update, it was only the packaging generated *here* that finally got fixed. Sure go for it...

alerque (approved):

bschwind (approved):
LGTM!

mcginty:
I definitely see the usefulness in keeping a warning around, but since we're just checking `argv` here any user-defined symlinks or aliases would trigger it, which would get really annoying.

I'd imagine our options are to remove it like this, or at least add a new `--suppress-inn-warning` type argument. Personally, agree I think the first is fine at this point.

strohel:
Thanks everyone for the quick discussion iterations. I'll go ahead and merge this as-is.

Let the `inn` symlink rest in peace. :headstone:

PR #268, opened 2023-06-14
update release.toml for latest cargo-release compat

mcginty:
`release.sh` script will fail otherwise. I honestly don't remember what `dev-version` was for, but a quick run of the script seems to do the right thing still.


bschwind (approved):
LGTM!

strohel (approved):

PR #272, opened 2023-06-14
Drop the release-artifacts (.deb packages) CI job

strohel:
Follow-up to #203. In that issue we've found out that the built packages don't work correctly on some Debian versions.

We've (@mcginty I think?) then manually removed them from the 1.5.4 release artifacts, and they have been provided by 3rd party repositories since.

So drop the CI jobs.


bschwind (approved):
LGTM!

PR #273, opened 2023-06-24
hostsfile: Copy the SELinux context to the temp file before overwrite

refi64:
On SELinux-enabled systems, /etc/hosts has a different type `net_conf_t` than the other files in /etc, so the temporary file that overwrites it ends up with the wrong context, resulting in many system services becoming unable to access the file. To fix this, manually look up the context /etc/hosts has and copy it to the temporary file before the rename.

In order to avoid depending on libselinux on systems that don't use it, this support is gated behind the new "selinux" feature. It *is* installed and enabled in the Dockerfile, however, in order to ensure that it still builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,19 +226,17 @@ Note that you'll be responsible for updating manually.
 
 ## Development
 
+### Cargo build feature for SELinux
+
+If your target system uses SELinux, you will want to enable the 'selinux' feature when building the innernet binary.
+This will ensure that innernet maintains the correct selinux context on the /etc/hosts file when adding hosts.  To do so add ```--features selinux``` to the ```cargo build``` options.
+The `selinux-devel` package will need to be installed for the correct headers.
+
 ### `innernet-server` Build dependencies
 
 - `rustc` / `cargo` (version 1.50.0 or higher)
 - `libclang` (see more info at [https://crates.io/crates/clang-sys](https://crates.io/crates/clang-sys))
 - `libsqlite3`
-
-### Cargo build feature for SELinux
-
-If your target system uses SELinux, you will want to enable the
-'selinux' feature when building the innernet binary. This will ensure
-that innernet maintains the correct selinux context on the /etc/hosts
-file when adding hosts.  To do so add ```--features selinux``` to the
-```cargo build``` options.
 
 Build:
 
--- a/README.md
+++ b/README.md
@@ -232,6 +232,14 @@ Note that you'll be responsible for updating manually.
 - `libclang` (see more info at [https://crates.io/crates/clang-sys](https://crates.io/crates/clang-sys))
 - `libsqlite3`
 
+### Cargo build feature for SELinux
+
+If your target system uses SELinux, you will want to enable the
+'selinux' feature when building the innernet binary. This will ensure
+that innernet maintains the correct selinux context on the /etc/hosts
+file when adding hosts.  To do so add ```--features selinux``` to the
+```cargo build``` options.
+
 Build:
 
 ```sh
--- a/hostsfile/src/lib.rs
+++ b/hostsfile/src/lib.rs
@@ -285,6 +285,48 @@ impl HostsBuilder {
     fn write_and_swap(temp_path: &Path, hosts_path: &Path, contents: &[u8]) -> io::Result<()> {
         // Copy the file we plan on modifying so its permissions and metadata are preserved.
         std::fs::copy(hosts_path, temp_path)?;
+
+        #[cfg(feature = "selinux")]
+        if selinux::current_mode() != selinux::SELinuxMode::NotRunning {
+            log::trace!("SELinux is running; copying context");
+            use selinux::SecurityContext;
+
+            const FOLLOW_SYMBOLIC_LINKS: bool = false;
+            const RAW_FORMAT: bool = false;
+            match SecurityContext::of_path(hosts_path, FOLLOW_SYMBOLIC_LINKS, RAW_FORMAT) {
+                Ok(Some(context)) => {
+                    log::trace!(
+                        "{} context is {:?}",
+                        hosts_path.display(),
+                        context.to_c_string()
+                    );
+                    if let Err(err) =
+                        context.set_for_path(temp_path, FOLLOW_SYMBOLIC_LINKS, RAW_FORMAT)
+                    {
+                        log::warn!(
+                            "SELinux context of {} ({:?}) could not be set \
+                            ({} may become inaccessible due to permission errors): {:?}",
+                            temp_path.display(),
+                            context.to_c_string(),
+                            hosts_path.display(),
+                            err
+                        );
+                    }
+                },
+                Ok(None) => {
+                    log::trace!("Hosts file {} had no SELinux context", hosts_path.display());
+                },
+                Err(err) => {
+                    log::warn!(
+                        "SELinux context of {} could not be retrieved \
+                        (file may become inaccessible due to permission errors): {:?}",
+                        hosts_path.display(),
+                        err
+                    );
+                },
+            }
+        }
+
         Self::write_clobber(temp_path, contents)?;
         std::fs::rename(temp_path, hosts_path)?;
         Ok(())
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -143,6 +143,29 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
+name = "bindgen"
+version = "0.69.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a00dc851838a2120612785d195287475a3ac45514741da670b735818822129a0"
+dependencies = [
+ "bitflags 2.5.0",
+ "cexpr",
+ "clang-sys",
+ "itertools",
+ "lazy_static",
+ "lazycell",
+ "log",
+ "prettyplease",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash",
+ "shlex",
+ "syn",
+ "which",
+]
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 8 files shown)


refi64:
hmm the CI failures seem unrelated?

```
[#] docker start 2bc1b0ae0db1e203e180da06a86cd383b8b76472feeead435e2cc35f216f8023
2bc1b0ae0db1e203e180da06a86cd3[83](https://github.com/tonarino/innernet/actions/runs/5361835266/jobs/9728189885?pr=273#step:4:84)b8b76472feeead435e2cc35f216f8023
 DEBUG wireguard_control::backends::kernel > get_by_name: got 1 response message(s) from netlink request
 DEBUG wireguard_control::backends::kernel > get_by_name: parsed wireguard device evilcorp with 1 peer(s)
- Creating a new CIDR from first peer.
[#] docker exec 2bc1b0ae0db1e203e180da06a[86](https://github.com/tonarino/innernet/actions/runs/5361835266/jobs/9728189885?pr=273#step:4:87)cd383b8b76472feeead435e2cc35f216f8023 innernet add-cidr evilcorp --name robots --cidr 10.66.2.0/24 --parent evilcorp --yes
Error: No such container: 2bc1b0ae0db1e203e180da06a86cd383b8b76472feeead435e2cc35f216f8023
```

but I don't think I have perms to retry.

strohel:
> but I don't think I have perms to retry.

Just retried them, let's see.

hostsfile/src/lib.rs
```diff
@@ -284,6 +284,44 @@ impl HostsBuilder {
...
+            let hosts = File::open(hosts_path)?;
+            let temp = File::options().append(true).open(temp_path)?;
+            match SecurityContext::of_file(&hosts.as_raw_fd(), false) {
+                Ok(Some(context)) => {
+                    log::debug!(
+                        "{} context is {:?}",
+                        hosts_path.display(),
+                        context.to_c_string()
+                    );
+                    if let Err(err) = context.set_for_file(&temp.as_raw_fd()) {
```
mcginty, line 303: Can we use [`of_path`](https://docs.rs/selinux/latest/selinux/struct.SecurityContext.html#method.of_path) and [`set_for_path`](https://docs.rs/selinux/latest/selinux/struct.SecurityContext.html#method.set_for_path) here?

```suggestion
            match SecurityContext::of_path(hosts_path, false) {
                Ok(Some(context)) => {
                    log::debug!(
                        "{} context is {:?}",
                        hosts_path.display(),
                        context.to_c_string()
                    );
                    if let Err(err) = context.set_for_path(temp_path) {
```

mcginty (commented):
Unfortunately I'm not an SELinux user so I can't really confirm the correctness of the logic, but the code looks clean enough.

Do you think it's worth the hassle to do this context copying compared to, say, only allowing overwriting `/etc/hosts` in-place if SELinux is detected?

hostsfile/src/lib.rs
refi64: ah didn't notice those functions were available, done!

refi64:
> Do you think it's worth the hassle to do this context copying compared to, say, only allowing overwriting /etc/hosts in-place if SELinux is detected?

IMO yes: SELinux-enabled is the default on a lot of systems, namely the RHEL variants & Fedora. Losing atomic updates on those systems in favor of potentially wiping the hosts file (learned the importance of atomic file updates the hard way before :upside_down_face:) would be pretty unfortunate :sweat_smile: .

mcginty (approved):
This seems reasonable to me - @strohel @bschwind any objections?

strohel (changes requested):
This looks good to me, I have just 2 considerations:
- @refi64 can you please rebase on current `main`? That should fix the Docker CI as a side-effect
- I guess selinux is specific to Linux, i.e. what happens if one tries to build innernet on a Mac with `selinux` feature enabled? Maybe we should make it target-specific, but I'm not sure how cargo features and target-specific dependencies interact. (ideally we don't have the feature at all outside linux, but I fear that's not possible)

hostsfile/src/lib.rs
```diff
@@ -284,6 +284,41 @@ impl HostsBuilder {
     fn write_and_swap(temp_path: &Path, hosts_path: &Path, contents: &[u8]) -> io::Result<()> {
         // Copy the file we plan on modifying so its permissions and metadata are preserved.
         std::fs::copy(hosts_path, temp_path)?;
+
+        #[cfg(feature = "selinux")]
+        if selinux::current_mode() != selinux::SELinuxMode::NotRunning {
+            log::debug!("SELinux is running; copying context");
```
strohel, line 290: Perhaps this could be demoted to `trace!`

hostsfile/src/lib.rs
```diff
@@ -284,6 +284,41 @@ impl HostsBuilder {
...
+            use selinux::SecurityContext;
+
+            match SecurityContext::of_path(hosts_path, false, false) {
+                Ok(Some(context)) => {
+                    log::debug!(
+                        "{} context is {:?}",
+                        hosts_path.display(),
+                        context.to_c_string()
+                    );
```
strohel, line 299: Also could be a trace I think.

hostsfile/src/lib.rs
```diff
@@ -284,6 +284,41 @@ impl HostsBuilder {
...
+                    if let Err(err) = context.set_for_path(temp_path, false, false) {
+                        log::warn!(
+                            "Failed to set selinux context of {} to {:?}: {:?}",
+                            temp_path.display(),
+                            context.to_c_string(),
+                            err
+                        );
```
strohel, line 306: I like when warnings like this also state the consequence. Here we could say something like "selinux context of {} was not preserved over atomic write"..

bschwind (approved):
Looks good, just a small suggestion.

I also agree with @strohel we need a rebase on `main` which should fix CI and the existing merge conflict.

hostsfile/src/lib.rs
```diff
@@ -284,6 +284,41 @@ impl HostsBuilder {
...
+                    }
+                },
+                Ok(None) => {
+                    log::debug!("Hosts file {} had no context", hosts_path.display());
```
strohel, line 310: Shall we mention "selinux" context? And ideally also the consequence, or demote this to trace altogether.

strohel (commented):
Adding some more notes on logging messages.

refi64:
whoops looks like this fell off my radar, all comments should be fixed now.

refi64:
Clippy failure seems unrelated:

```
error: accessing first element with `responses.get(0)`
Error:   --> netlink-request/src/lib.rs:64:19
   |
64 |             match responses.get(0) {
   |                   ^^^^^^^^^^^^^^^^ help: try: `responses.first()`
```

bschwind:
@refi64 I'll take care of the clippy issues in `main` and then we can rebase this on top of those changes.

bschwind:
@refi64 done, if you rebase on `main` the clippy step should now pass.

refi64:
@bschwind just rebased :crossed_fingers:

bschwind (approved):
LGTM++

I'll wait for a quick review from @strohel before merging.

strohel (approved):
This is looking good, but one late request @refi64: could you please document the new feature (flag) in README.md? Also stating that it is currently disabled by default. We'll then make sure we mention it in the next relase notes, pinging package maintainers to decide whether they want to activate the flag or not.

jebotz:
> This is looking good, but one late request @refi64: could you please document the new feature (flag) in README.md? Also stating that it is currently disabled by default. 

Since @refi64 hasn't responded, and I'd really like to see this merged (I use the code, works fine for me) I've added the docemntation; I'm including it here as a patch...
```
From 831d19d907e004e89dde797d9191b07700c78d77 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?J=C3=BCrgen=20Botz?= <jurgen@botz.org>
Date: Sun, 28 Apr 2024 17:02:13 +0200
Subject: [PATCH] Add info about selinux feature to README.md

---
 README.md | 8 ++++++++
 1 file changed, 8 insertions(+)

diff --git a/README.md b/README.md
index b6ee618..592afc1 100644
--- a/README.md
+++ b/README.md
@@ -232,6 +232,14 @@ Note that you'll be responsible for updating manually.
 - `libclang` (see more info at [https://crates.io/crates/clang-sys](https://crates.io/crates/clang-sys))
 - `libsqlite3`
 
+### Cargo build feature for SELinux
+
+If your target system uses SELinux, you will want to enable the
+'selinux' feature when building the innernet binary. This will ensure
+that innernet maintains the correct selinux context on the /etc/hosts
+file when adding hosts.  To do so add ```--features selinux``` to the
+```cargo build``` options.
+
 Build:
 
 ```sh
-- 
2.44.0

```

bschwind:
Thanks, @jebotz! Thankfully @refi64 has allowed (perhaps by default) project maintainers to push to fork branches which back a PR on the main repo, so I was able to update it and incorporate your change. I'll try to get this merged once I fix the CI errors.

refi64:
oops sorry it looks like I completely missed the last comment :sweat_smile: thanks for pushing that update!

README.md
```diff
@@ -232,6 +232,14 @@ Note that you'll be responsible for updating manually.
...
+
+If your target system uses SELinux, you will want to enable the
+'selinux' feature when building the innernet binary. This will ensure
+that innernet maintains the correct selinux context on the /etc/hosts
+file when adding hosts.  To do so add ```--features selinux``` to the
+```cargo build``` options.
+
 Build:
 
 ```sh
```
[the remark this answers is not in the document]
jebotz: OK, here's another patch with those changes.  I also remembered that the build host will need the selinux headers from libselinux-devel and added a comment to that effect.

```

From cb9d0735ab3681bc46fe902a770753be3b3aec81 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?J=C3=BCrgen=20Botz?= <jurgen@botz.org>
Date: Tue, 7 May 2024 06:41:06 +0200
Subject: [PATCH] Reformatted & repositioned and improved doc about selinux
 feature

---
 README.md | 14 ++++++--------
 1 file changed, 6 insertions(+), 8 deletions(-)

diff --git a/README.md b/README.md
index aa9cec3..272c978 100644
--- a/README.md
+++ b/README.md
@@ -226,20 +226,18 @@ Note that you'll be responsible for updating manually.
 
 ## Development
 
+### Cargo build feature for SELinux
+
+If your target system uses SELinux, you will want to enable the 'selinux' feature when building the innernet binary.
+This will ensure that innernet maintains the correct selinux context on the /etc/hosts file when adding hosts.  To do so add ```--features selinux``` to the ```cargo build``` options.
+The `selinux-devel` package will need to be installed for the correct headers.
+
 ### `innernet-server` Build dependencies
 
 - `rustc` / `cargo` (version 1.50.0 or higher)
 - `libclang` (see more info at [https://crates.io/crates/clang-sys](https://crates.io/crates/clang-sys))
 - `libsqlite3`
 
-### Cargo build feature for SELinux
-
-If your target system uses SELinux, you will want to enable the
-'selinux' feature when building the innernet binary. This will ensure
-that innernet maintains the correct selinux context on the /etc/hosts
-file when adding hosts.  To do so add ```--features selinux``` to the
-```cargo build``` options.
-
 Build:
 
 ```sh
-- 
2.45.0
```

jebotz:
So, how about someone merge this?

README.md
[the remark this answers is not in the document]
bschwind: Thanks @jebotz, I applied the patch. Sorry for the long wait on this one!

PR #275, opened 2023-07-07
publicip: fix IPv4 resolution on macOS

mcginty:
macOS isn't happy with using the IPv6 unspecified address when routing over the IPv4 stack.


strohel (approved):
LGTM

bschwind:
Wondering why this is now failing in the docker tests...is it because we're now binding to an IPV4 address instead of IPV6?

strohel:
@bschwind Dockers tests are failing also in master for some time, so that should be independent. (but yeah we should have a look)

bschwind (approved):
Ah, didn't realize the tests are failing in `main`. Weird that they started after we released 1.60 and essentially just changed some shell completion files...

Anyway, this code looks good!

Issue #277, opened 2023-07-08
rpm doesn't build for CentOS 7


wwalker:
On CentOS 7 rpm doesnt' build.

2 reasons:
* `cc` is missing
* `dnf` doesn't exist on CentOS 7 (it still has `yum`, `dnf`'s predecessor)

PR in just a minute.

PR #278, opened 2023-07-08
#277 - add gcc, change dnf to yum

wwalker:
* add the `gcc` package to the packages to add to the docker image for build
* change `dnf` to `yum`.  (systems that use `dnf` always have `yum` and the basic `yum/dnf` commands are identical)
* I have built packages for Fedora 37,38 and CentOS 7 with this fix applied

### DIFF
--- a/rpm/Dockerfile
+++ b/rpm/Dockerfile
@@ -4,8 +4,8 @@ ARG VER
 FROM ${DISTRO}:${VER} as builder
 LABEL stage=innernet-rpm
 
-RUN dnf -y update && \
-	dnf -y install clang-devel sqlite-devel glibc-devel rpm-build && \
+RUN yum -y update && \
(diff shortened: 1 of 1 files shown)


mcginty (approved):
Tested on Alma also for sanity with modern redhat derivatives and seems fine!

strohel (approved):
Also LGTM. With .deb package we've actually transitoned away from providing them within our repository - an external build by a community member is now provided itself.

With RPMs we probably don't need to remove what's working, but I wonder @wwalker if you'd be interesting in maintaining an external RPM build/repository for innernet?

wwalker:
I can probably do that.  I'll get back on this in the next two days.

strohel:
> I can probably do that. I'll get back on this in the next two days.

:pray: :pray: :pray:  In the mean time I've merged this.

PR #282, opened 2023-08-22
Fix docker tests

mcginty:
At some point, the golang and rustlang docker images decided that `latest` meant different things. Golang was building on the new `bookworm` Debian version, while Rust wasn't. This forces them both to a specific version of Debian to avoid glibc conflicts.


bschwind (approved):
LGTM! Good to have the nicer test output too.

strohel (approved):
Oh thanks for fixing the CI @mcginty! Yay!

PR #284, opened 2023-08-26
 Add Error Message when wireguard-go not found. replacement for #232

noyez:
Previously when trying to add a peer, i for the following output, which
was not helpful since i wasn't sure what file could not be found.
```
bkn@mbp16 bkn % sudo /Users/bkn/.cargo/bin/innernet  --verbose  install mbp16.toml
✔ Interface name · chester
[*] bringing up interface chester.
[E] failed to start the interface: chester - No such file or directory (os error 2).
[*] bringing down the interface.
[!] failed to bring down interface: chester - WireGuard name file can't be read.
[E] Failed to redeem invite. Now's a good time to make sure the server is started and accessible!

[E] chester - No such file or directory (os error 2)
```

After changes: 
```
[1] bkn@mbp16> sudo target/debug/innernet --verbose  install /Users/brad/mbp16.conf             /Users/brad/source/rust/innernet
✔ Interface name · chester2
[*] bringing up interface chester2.
[E] failed to start the interface: chester2 - Cannot find "wireguard-go". Specify a custom path with WG_USERSPACE_IMPLEMENTATION..
[*] bringing down the interface.
[!] failed to bring down interface: chester2 - WireGuard name file can't be read.
[E] Failed to redeem invite. Now's a good time to make sure the server is started and accessible!

[E] chester2 - Cannot find "wireguard-go". Specify a custom path with WG_USERSPACE_IMPLEMENTATION.
```

ref  new PR to replace #232

### DIFF
--- a/wireguard-control/src/backends/userspace.rs
+++ b/wireguard-control/src/backends/userspace.rs
@@ -274,6 +274,34 @@ fn get_userspace_implementation() -> String {
 
 fn start_userspace_wireguard(iface: &InterfaceName) -> io::Result<Output> {
     let mut command = Command::new(get_userspace_implementation());
+
+    let output_res = if cfg!(target_os = "linux") {
+        command.args(&[iface.to_string()]).output()
+    } else {
+        command
+            .env("WG_TUN_NAME_FILE", &format!("{VAR_RUN_PATH}/{iface}.name"))
+            .args(["utun"])
+            .output()
+    };
+
+    match output_res {
+        Ok(output) => {
+            if output.status.success() {
+                Ok(output)
+            } else {
+                Err(io::ErrorKind::AddrNotAvailable.into())
+            }
+        },
+        Err(e) if e.kind() == io::ErrorKind::NotFound => Err(io::Error::new(
+            io::ErrorKind::NotFound,
+            "Cannot find \"wireguard-go\". Specify a custom path with WG_USERSPACE_IMPLEMENTATION.",
+        )),
+        Err(e) => Err(e),
+    }
+}
+
+fn _start_userspace_wireguard(iface: &InterfaceName) -> io::Result<Output> {
+    let mut command = Command::new(get_userspace_implementation());
     let output = if cfg!(target_os = "linux") {
         command.args(&[iface.to_string()]).output()?
     } else {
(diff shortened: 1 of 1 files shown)


noyez:
On a semi-related topic you might also consider adding functions to output the env-vars of WG_USERSPACE_IMPLEMENTATION, like so: 
```
fn wg_userspace_implementation_env_var() -> &'static str { "WG_USERSPACE_IMPLEMENTATION" }
fn wg_quick_userspace_implementation_env_var() -> &'static str { "WG_QUICK_USERSPACE_IMPLEMENTATION" }

```
this will eliminate the same hardcoded strings in multiple locations throughout the code base. As someone who is prone to typos, this i had misspelled one of those env vars and it stumped my testing for a bit, so i thought why not just put those in a function (or a `const` or a `const fn`) or something.

skywhale (approved):
LGTM

I'm sorry that the PR is left unreviewed for a long time 🙇

wireguard-control/src/backends/userspace.rs
```diff
@@ -274,18 +274,29 @@ fn get_userspace_implementation() -> String {
...
+        Ok(output) => {
+            if output.status.success() {
+                Ok(output)
+            } else {
+                Err(io::ErrorKind::AddrNotAvailable.into())
+            }
+        },
+        Err(e) if e.kind() == io::ErrorKind::NotFound => Err(io::Error::new(
+            io::ErrorKind::NotFound,
+            "Cannot find \"wireguard-go\". Specify a custom path with WG_USERSPACE_IMPLEMENTATION.",
```
bschwind, line 297: We can probably pull `get_userspace_implementation()` into a local var called `userspace_implementation` and then use it in `Command::new()` above, as well as here:

```suggestion
            "Cannot find {userspace_implementation:?}. Specify a custom path with WG_USERSPACE_IMPLEMENTATION.",
```

bschwind (approved):
Also sorry for an incredibly late review on this, but it looks good! I just have on minor point to address.

goodhoko (approved):
LGTM! Thanks a lot @noyez and forgive these late responses.

I've pushed a commit to address @bschwind review. So this should be good to go I believe!

wireguard-control/src/backends/userspace.rs
goodhoko: Addressed in the latest commit.

Issue #309, opened 2024-04-18
Add `innernet rename-cidr`


strohel:
We already support renaming peers since https://github.com/tonarino/innernet/pull/92.
CIDRs can also be deleted thanks to https://github.com/tonarino/innernet/pull/88.

But we don't have the ability to rename CIDRs (short of editing the server database). This issue tracks adding it.

PR #310, opened 2024-04-18
Add a new client / server command to rename CIDR

skywhale:
Closes https://github.com/tonarino/innernet/issues/309

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -754,10 +754,9 @@ fn rename_cidr(
 
         let id = cidrs
             .iter()
-            .filter(|c| c.name == old_name)
-            .map(|c| c.id)
-            .next()
-            .ok_or_else(|| anyhow!("CIDR not found."))?;
+            .find(|c| c.name == old_name)
+            .ok_or_else(|| anyhow!("CIDR not found."))?
+            .id;
 
         api.http_form("PUT", &format!("/admin/cidrs/{id}"), cidr_request)?;
         log::info!("CIDR renamed.");
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -12,7 +12,7 @@ use shared::{
     AddCidrOpts, AddDeleteAssociationOpts, AddPeerOpts, Association, AssociationContents, Cidr,
     CidrTree, DeleteCidrOpts, EnableDisablePeerOpts, Endpoint, EndpointContents, InstallOpts,
     Interface, IoErrorContext, ListenPortOpts, NatOpts, NetworkOpts, OverrideEndpointOpts, Peer,
-    RedeemContents, RenamePeerOpts, State, WrappedIoError, REDEEM_TRANSITION_WAIT,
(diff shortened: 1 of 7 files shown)


server/src/db/cidr.rs
```diff
@@ -106,6 +112,22 @@ impl DatabaseCidr {
         Ok(Cidr { id, contents })
     }
 
+    /// Update self with new contents, validating them and updating the backend in the process.
+    pub fn update(&mut self, conn: &Connection, contents: CidrContents) -> Result<(), ServerError> {
```
strohel, line 116:
```suggestion
    /// Update self with new contents, validating them and updating the backend in the process.
    /// Currently this only supports updating the name and ignores changes to any other field.
    pub fn update(&mut self, conn: &Connection, contents: CidrContents) -> Result<(), ServerError> {
```

client/src/main.rs
```diff
@@ -724,6 +737,37 @@ fn add_cidr(interface: &InterfaceName, opts: &Opts, sub_opts: AddCidrOpts) -> Re
...
+
+    if let Some((cidr_request, old_name)) = prompts::rename_cidr(&cidrs, &sub_opts)? {
+        log::info!("Renaming CIDR...");
+
+        let id = cidrs
+            .iter()
+            .filter(|c| c.name == old_name)
+            .map(|c| c.id)
+            .next()
+            .ok_or_else(|| anyhow!("CIDR not found."))?;
```
strohel, line 760: I've tested this compiles:
```suggestion
        let id = cidrs
            .iter()
            .find(|c| c.name == old_name)
            .ok_or_else(|| anyhow!("CIDR not found."))?
            .id;
```

strohel (approved):
Oh great! Adding just 2 little technicalities, but approving already.

I was about to ask about making sure that CIDR names are unique, but I see the database takes care of it (same as in peer names), which is good, as that's probably the only non-racy way to check it.

server/src/main.rs
```diff
@@ -460,6 +470,27 @@ fn add_cidr(
...
+    let cidrs = DatabaseCidr::list(&conn)?;
+
+    if let Some((cidr_request, old_name)) = shared::prompts::rename_cidr(&cidrs, &opts)? {
+        let db_cidr = DatabaseCidr::list(&conn)?
+            .into_iter()
+            .find(|c| c.name == old_name)
+            .ok_or_else(|| anyhow!("CIDR not found."))?;
+        db::DatabaseCidr::from(db_cidr).update(&conn, cidr_request)?;
+    } else {
+        println!("exited without creating CIDR.");
```
mcginty, line 488:
```suggestion
        println!("exited without renaming CIDR.");
```

mcginty (commented):
Looks good! One small copy-pasta typo fix.

skywhale:
@strohel @mcginty thank you for your review :)

I commited the suggested changes and checked one more time that docker tests succeed for the new test case.

strohel (approved):
LGTM++

mcginty (approved):
🙃

PR #311, opened 2024-04-18
Make docker-tests less verbose and easier to add new tests

skywhale:


strohel (approved):
Oh thanks! Looked good, though my eye is not trained to spot problems in bash.

But this is test infra, so not worth to delay this by asking more eyes on this, though if @bschwind (or @mcginty) wants to have a (post-merge) look I'd certainly wouldn't protest.

mcginty (approved):
I haven't tested it but this looks so much cleaner and more maintainable, great stuff.

Issue #308, opened 2024-04-11
Add some form of override-endpoint-port functionality


bschwind:
Someone chime in if this doesn't make sense, but while we have the option to override an _endpoint_ (IP + port), we don't have the ability to override just the port.

This could be nice for machines that don't have a stable IP, but have port forwarding configured on their router.

skywhale:
where do we take the IP from? do we already have a mechanism to query outside to get a global IP?

maybe `override-endpoint-port` to avoid confusion with the local listen port.

bschwind:
When we run `override-endpoint`, we query 1.1.1.1 to get the external endpoint.

The innernet server also takes note of the IP used by the client to connect to the server.

And yeah `override-endpoint-port` or something similar makes sense to avoid confusion with the listening port.

PR #315, opened 2024-06-14
Support override-endpoint with the unspecified address

skywhale:
An alternative way to implement https://github.com/tonarino/innernet/issues/308

What do you think @bschwind?

### DIFF
--- a/shared/src/prompts.rs
+++ b/shared/src/prompts.rs
@@ -15,7 +15,7 @@ use std::{
     fmt::{Debug, Display},
     fs::{File, OpenOptions},
     io,
-    net::SocketAddr,
+    net::{IpAddr, Ipv6Addr, SocketAddr},
     str::FromStr,
     time::SystemTime,
 };
@@ -565,6 +565,12 @@ pub fn ask_endpoint(listen_port: u16) -> Result<Endpoint, Error> {
         .interact()?
     {
         publicip::get_any(Preference::Ipv4)
+    } else if Confirm::with_theme(&*THEME)
+        .wait_for_newline(true)
+        .with_prompt("You do not have a fixed global IP (use the unspecified address)?")
+        .interact()?
+    {
+        Some(IpAddr::V6(Ipv6Addr::UNSPECIFIED))
     } else {
         None
     };
--- a/shared/src/types.rs
+++ b/shared/src/types.rs
@@ -146,6 +146,19 @@ impl Endpoint {
             )
         })
     }
+
+    /// Returns true if the endpoint host is unspecified e.g. 0.0.0.0
+    pub fn is_host_unspecified(&self) -> bool {
+        match self.host {
+            Host::Ipv4(ip) => ip.is_unspecified(),
+            Host::Ipv6(ip) => ip.is_unspecified(),
+            Host::Domain(_) => false,
+        }
+    }
+
+    pub fn port(&self) -> u16 {
+        self.port
+    }
 }
 
 #[derive(Deserialize, Serialize, Debug)]
@@ -437,7 +450,9 @@ pub struct ListenPortOpts {
 
 #[derive(Debug, Clone, PartialEq, Eq, Args)]
 pub struct OverrideEndpointOpts {
-    /// The listen port you'd like to set for the interface
+    /// The external endpoint that you'd like the innernet server to broadcast
+    /// to other peers. The IP address may be unspecified, in which case the
+    /// server will try to resolve it based on its most recent connection.
     #[clap(short, long)]
     pub endpoint: Option<Endpoint>,
 
--- a/server/src/api/mod.rs
+++ b/server/src/api/mod.rs
@@ -1,26 +1,40 @@
-use shared::Peer;
+use std::net::SocketAddr;
 
 use crate::Session;
+use shared::{Endpoint, Peer};
 
 pub mod admin;
 pub mod user;
 
-/// Inject the collected endpoints from the WG interface into a list of peers.
-/// This is essentially what adds NAT holepunching functionality. If a peer
-/// already has an endpoint specified (by calling the override-endpoint) API,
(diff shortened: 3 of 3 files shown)


client/src/main.rs
```diff
@@ -281,22 +281,6 @@ enum Command {
     },
 }
 
-/// Application-level error.
```
skywhale, line 284: This was a clippy warning from the lastest 1.78.0

bschwind: What was the clippy warning?

shared/src/prompts.rs
```diff
@@ -565,6 +565,12 @@ pub fn ask_endpoint(listen_port: u16) -> Result<Endpoint, Error> {
         .interact()?
     {
         publicip::get_any(Preference::Ipv4)
+    } else if Confirm::with_theme(&*THEME)
+        .wait_for_newline(true)
+        .with_prompt("You do not have a fixed global IP (use the unspecified address)?")
```
bschwind, line 570: This doesn't quite read like a question to me, possible re-wording:

```suggestion
        .with_prompt("Use an unspecified IP address? (this can occur if you do not have a fixed global IP)")
```

shared/src/types.rs
```diff
@@ -437,7 +450,9 @@ pub struct ListenPortOpts {
 
 #[derive(Debug, Clone, PartialEq, Eq, Args)]
 pub struct OverrideEndpointOpts {
-    /// The listen port you'd like to set for the interface
+    /// The external endpoint that you'd like the innernet server to broadcast
+    /// to other peers. The IP address may be unspecified, in which case the
```
bschwind, line 454: Should we mention that the port will still be used, even if you decide to use an unspecified IP address?

bschwind (commented):
Nice, I think this could work as an alternative! I'd like @strohel or @mcginty to double check it too, as there might be edge cases I'm missing.

client/src/main.rs
strohel:
```
warning: struct `ClientError` is never constructed
   --> client/src/main.rs:286:19
    |
286 | pub(crate) struct ClientError(String);
    |                   ^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

skywhale:
Thank you @bschwind :)

@strohel What do you think? I still need to / want to add a docker test, but it'd be helpful to get your initial impression.

shared/src/prompts.rs
```diff
@@ -565,6 +565,14 @@ pub fn ask_endpoint(listen_port: u16) -> Result<Endpoint, Error> {
         .interact()?
     {
         publicip::get_any(Preference::Ipv4)
+    } else if Confirm::with_theme(&*THEME)
+        .wait_for_newline(true)
+        .with_prompt(
+            "Use an unspecified IP address? (this can occur if you do not have a fixed global IP)",
```
strohel, line 571:
```suggestion
            "Use an unspecified IP address and override just the port? (use if you do not have \
             a fixed global IP)",
```

shared/src/types.rs
```diff
@@ -437,7 +450,11 @@ pub struct ListenPortOpts {
...
+    /// server will try to resolve it based on its most recent connection.
+    /// The port will still be used even if you decide to use the unspecified
+    /// IP address.
     #[clap(short, long)]
     pub endpoint: Option<Endpoint>,
```
strohel, line 459: Before this I didn't even know what "unspecified IP address" exactly means. Shall we give an example like `0.0.0.0:1234`?

shared/src/prompts.rs
```diff
@@ -565,6 +565,14 @@ pub fn ask_endpoint(listen_port: u16) -> Result<Endpoint, Error> {
...
+        )
+        .interact()?
+    {
+        Some(IpAddr::V6(Ipv6Addr::UNSPECIFIED))
```
strohel, line 575: I guess it does not really make a difference whether the unspecified IP address is IPv4 or IPv6? Still, we generally default to IPv4, so we may as well follow suit here.

strohel (changes requested):
Somehow it was pretty challenging for my mind to review this. I think this is a good direction in general, but I suggest some slight changes to the server logic.

In addition to inline comments, I think we should also consider backwards and forwards compatibility:
1. new server, old client scenario: should work just fine; even one new client reporting unspecified addr while other clients being old works well, the server "resolves" the endpoint for them.
2. new client, old server: then I think the server ends up sending address like 0.0.0.0:1234 to clients. I'd say this is still acceptable if it doesn't break connections - i.e. if https://github.com/tonarino/innernet/pull/305 still saves the day in that case. But we could check the server version on the client and refuse to use unspecified address if it is old, if that's easy.

   We could also filter our the unspecified addresses from peer endpoints/candidates on client side for this scenario when trying to connect to other peers.

strohel:
@skywhale pingy ping in case you still have cycles to iterate this PR.

bschwind (approved):
Looking good! I agree with all of @strohel's suggestions and this should be good to merge after they're applied.

mbernat:
@strohel @bschwind The comments should be resolved, PTAL 🙏 

Regarding compatibility with old servers.

> new client, old server: then I think the server ends up sending address like 0.0.0.0:1234 to clients. I'd say this is still acceptable if it doesn't break connections - i.e. if https://github.com/tonarino/innernet/pull/305 still saves the day in that case. But we could check the server version on the client and refuse to use unspecified address if it is old, if that's easy.

Let me try to add the version check.

> We could also filter our the unspecified addresses from peer endpoints/candidates on client side for this scenario when trying to connect to other peers.

Client filtering only helps on the new clients. If there are also old clients on the network they'd still receive and use unspecified addresses from the new clients. So, this approach is less appealing.

strohel (approved):
Looking good! Feel free to merge.

Issue #11, opened 2021-03-30
Consider publishing to crates.io


jhseu:
For those of us who prefer to cargo install and build from source rather than using a binary.

mcginty:
Absolutely - I'm planning on doing that after removing our git-based dependencies (crates.io doesn't allow publishing crates with them). Sorry we didn't mention that in the README as something that was in the works.

mcginty:
Note that in the mean time you can checkout the source and run `cargo install --path ./client` and `cargo install --path ./server` and it works fine.

pilik88:
👍

[pilik88 - Chat @ Spike](https://spikenow.com/r/a/?ref=spike-organic-signature&_ts=zd7re)	[zd7re]

On April 10, 2021 at 6:32 GMT, Jake McGinty ***@***.***> wrote:

Note that in the mean time you can checkout the source and run cargo install --path ./client and cargo install --path ./server and it works fine.

mediclab:
Can you update **wireguard-control** to 1.6.1 on crates.io ?

strohel:
> Can you update **wireguard-control** to 1.6.1 on crates.io ?

Sorry this a late reply: I've already done some preparations like https://github.com/tonarino/innernet/pull/321 and publishing `netlink-request` to crates.io, but I got stopped by the fact that only @mcginty is the owner of https://crates.io/crates/wireguard-control

Jake, could you please add `github:tonarino:engineering` (or alternatively select individuals) as co-owners of it? :pray:

strohel:
> Can you update **wireguard-control** to 1.6.1 on crates.io ?

Alright this took _way_ longer than anticipated, but it's done now.

The published 1.6.1 version doesn't exactly match the `v1.6.1` tag in this repo, but it has the same Rust code.

PR #320, opened 2024-08-27
Rename `server` to `innernet-server`, expose as a library

sqrtsanta:
This extracts commands from innernet-server binary so that it will be possible to use it as a library. This way it will be easier to make own wrappers around innernet-server e.g. gprc/http server. 

It exports certain commands:

- `add_cidr`
- `rename_cidr`
- `delete_cidr`
- `add_peer`
- `rename_peer`
- `enable_or_disable_peer`
- `serve`
- `uninstall`

It exports `initialize::init_wizard` as well, although my guess is this command will be used via CLI only.


sqrtsanta:
@strohel Hi! Have a couple of questions regarding this PR:

1. Would it makes sense to introduce `Server` struct, so that it won't be necessary to `open_database_connection` every time user uses any of server commands? It won't really change anything for a CLI, but will make it more convenient for library usage. 
2. What would be library name, `innernet_server`?
3. Other than docs, is there anything else I should cover?

server/src/lib.rs
```diff
@@ -0,0 +1,724 @@
...
+    });
+
+    let server = hyper::Server::from_tcp(listener)?.serve(make_svc);
+
+    server.await?;
+
+    Ok(())
+}
+
+fn spawn_endpoint_refresher(interface: InterfaceName, network: NetworkOpts) -> Endpoints {
```
strohel, line 481: I see this and spawn_expired_invite_sweeper were moved w.r.t. their original place in `main.rs` - could you please restore their original position above `async fn serve(`?

It will make the diff (when viewed with copy detection) smaller for free, thus easier to check.

strohel (commented):
Looking good, just one little meta-nit.

For the record I'm looking at the changes locally using
```
git fetch origin pull/320/head:extract-server-to-lib
git diff --find-copies main...extract-server-to-lib
```
which is able to detect the file move.

To answer your questions:
> 1. Would it makes sense to introduce `Server` struct, so that it won't be necessary to `open_database_connection` every time user uses any of server commands? It won't really change anything for a CLI, but will make it more convenient for library usage.

Yeah it would make sense, though I'd prefer for that to be done in a separate PR on top of this one (to be able so see the _actual_ diff) - reserving this one to be just the move.

> 2. What would be library name, `innernet_server`?

Yeah perhaps, maybe `innernet-server` for consistency. That's already used for the binary name, but that shuoldn't conflict?

> 3. Other than docs, is there anything else I should cover?

Nothing I could think of, maybe see if any part of the README can be updated.

I see the Docker tests CI fails, but if also fails when I ran in ton `main` now: https://github.com/tonarino/innernet/actions/runs/10316957718
It would be great to fix that separately and merge to main before this one.

---

I'd also love if either @bschwind or @mcginty could review :pray:.

server/src/lib.rs
sqrtsanta: Sure :+1:

sqrtsanta:
Hi, thanks for a review!

> Yeah perhaps, maybe innernet-server for consistency. That's already used for the binary name, but that shuoldn't conflict?

Unforunately, `innernet-server` as a library name is not possible due to `library target names cannot contain hyphens`. Also, right now package name is `server`, should I change it too, or leave it as `server` and only provide `lib.name`. I suspect it's a little confusing, when they don't match. 

> which is able to detect the file move.

Doesn't git marks it automatically based on content or should I somehow assist here?

sqrtsanta:
I've changed package's name to `innernet-server`. As for README.md, I'll leave it as it is now, since it won't be available on crates.io. I assume there will be changes to release.sh/etc, then it makes more sense to include updates to README.md in that PR.

server/src/main.rs
```diff
@@ -283,22 +140,32 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 std::process::exit(1);
             }
         },
-        Command::Uninstall { interface, yes } => uninstall(&interface, &conf, opts.network, yes)?,
+        Command::Uninstall { interface, yes } => {
+            innernet_server::uninstall(&interface, &conf, opts.network, yes)?
```
strohel, line 144: Let's import items from `innernet_server` instead of qualifying them. It will make the diff smaller (and lower the fallout from renaming to `innernet_server`).

strohel (commented):
Hi @sqrtsanta, sorry for the delay getting back to you!

One little suggestion (sorry only noticed now even though the code was also in your previous version), plus I'd still like to see a review from @bschwind or @mcginty. Or perhaps if you have some hints on the Docker CI failures on `main`? :pray:

> I've changed package's name to `innernet-server`.

That's good! Also a nice preparation towards https://github.com/tonarino/innernet/issues/11

> Unfortunately had to change installation instructions due to that and that made inconsistency between client & server:

I think this is fine. We may (have to) rename `client` to  `innernet-client` later anyway.

sqrtsanta:
Thanks @strohel! Docker CI failures are due to `golang.org/x/net/internal/socket` & Go1.23. Here https://github.com/bettercap/bettercap/issues/1106 folks stumbled upon the same problem. We either downgrade to Go1.22 or wait until wireguard-go [bumps](https://github.com/WireGuard/wireguard-go/blob/master/go.mod#L7) x/net.

bschwind (approved):
Hey this is great! Sorry I'm so late in reviewing it. I don't have much to add that @strohel hasn't already mentioned so just throwing my +1 in here.

Seems like we got a new clippy lint triggering after a toolchain upgrade. We could fix that here or in a separate PR to main. The dockerfile build will still fail either way though :(

strohel:
@sqrtsanta we've finally fixed the Docker tests issue (thanks for the hint!) and a new clippy issues that popped in the mean time.

Please rebase your PR on `main`, then the CI should go green again. Feel free squash your PR to a single commit first, as that may help with the rebase (I touched `main.rs` in the clippy PR).

sqrtsanta:
@strohel Done, should be good now.

PR #321, opened 2024-09-05
Preparations to publish wireguard-control, netlink-request

strohel:
### Change netlink-request license to MIT

Previous `LGPL-2.1-or-later` was a copy-paste error from wireguard-control. I unfortunately published it that way to creates.io...

### Update README with a note on publishing to crates.io


netlink-request/README.md
```diff
@@ -0,0 +1,5 @@
+# `netlink-request`
+
+A helper to perform requests using the Linux netlink interface.
+
+Versioning is held in lockstep with innernet, although this may change in the future.
```
goodhoko, line 5: IIUC, it would make sense to edit https://github.com/tonarino/innernet/blob/main/release.sh to handle `netlink-request/Cargo.toml` the same way as `wireguard-control/Cargo.toml` to ensure the version is locked in step.

mcginty: +1 to this being the easiest direction to take it.

As a future thought, these libraries are less married to the development of innernet directly, so there may be more cases in the future where you'd want to push a fix for `netlink-request` without needing to update the innernet semver at the same time.

mcginty (approved):
Besides @goodhoko's note about updating `release.sh`, looks good.

netlink-request/README.md
strohel: Good catch, added to `release.sh` in 3703001, I've manually checked that the `perl` command works.

On a slightly unrelated note, I think `cargo release` can now supply many duties we do manually in `release.sh` (replacing versions, committing, tagging). I see we have that disabled in `release.toml`. At some point we may revisit that.

goodhoko (approved):
ﾚムｲﾶ

PR #328, opened 2024-12-07
Fix README typo

wyager:
Component is called `server`


strohel (commented):
This is somwhat unfortunate situation: the crate was renamed to `innernet-server` in https://github.com/tonarino/innernet/pull/320 (part of main, but not part of the latest release 1.6.1).

So the readme _is_ incorrect now, but won't be as soon as we cut a new release.

Which is maybe the best resolution: let us cut a new release soon, making the readme correct.

Issue #335, opened 2025-03-02
can innernet-server node also become a member?


Oaklight:
I have deployed a innernet-server on my vps. but meanwhile I want to connect this vps to a cidr reserved for all my cloud machines. Can this be achievable? I tried create a peer toml for this server vps, but `innernet install xxx.toml` with neither original interface nor altered interface name works.

bschwind:
Hey @Oaklight, there are a couple issues related to what you're trying:

https://github.com/tonarino/innernet/issues/45

https://github.com/tonarino/innernet/issues/135

The short answer is that this is currently not supported, but given that your VPS is the innernet-server, it should be accessible to all peers on the innernet network.

Oaklight:
The attempt to include server as peer was out of the expectation to connect to peers from server. Because when I check the /etc/hosts from the server machine, the peers are not visible.

strohel:
@Oaklight oh, that was just beause the server doesn't currently write to `/etc/hosts`, but the peers are fully reachable form it using their innernet IPs.

Oaklight:
is writing to /etc/hosts for server machine something we can expect in the foreseeable future?
Also, how do we re-generate invites if something needs to be changed before its redeemed?

bschwind:
@Oaklight right, there's no command for it. I _believe_ you would have to delete the peer via an sqlite query, something like this:

```
DELETE FROM peers
WHERE is_redeemed = 0
```

You can of course tack on other conditions or just delete the exact peer directly, but I'll let you read up on the discussion on peer deletion and the rationale for making it not easy:

https://github.com/tonarino/innernet/issues/9

https://github.com/tonarino/innernet/issues/139

Sorry this probably isn't a great introduction to using innernet. Our typical workflow if we want to remove a peer is to disable the one we made and simply create a new one. I know if the CIDR you set up has a limited number of IPs this can be annoying, but just know you can always manually fix things up in sqlite (after taking a backup!) if needed.

alerque:
> > is writing to /etc/hosts for server machine something we can expect in the foreseeable future?
> 
> That should be pretty easy to do I would think, just mirror how the client does it already. We just haven't...done it yet.

Really, that's all that's needed? I've spent far more time working around this and setting up relay servers than it probably would have taken to just fix this. And that's just me, not every other user that has bumped up against this limitation!

I'll try to have a look at it myself and if it really that easy maybe I can contribute. In any case know that this would be a high-value addition.

[...]

Issue #135, opened 2021-08-27
Cannot redeem invite on same machine as server


russkel:
Hello,

I am trying to redeem an invite on the same machine as the server but I am getting a HTTP 401 error.

I had someone else redeem an invite on the server and it worked. Any ideas what's happening?

```
↳ inn --version
innernet 1.4.1

$ russ @ devmachine in ~ [18:26:31]
↳ sudo innernet-server add-peer testinnernet
✔ Eligible CIDRs for peer · devnet (10.42.1.0/24)
✔ IP · 10.42.1.4
✔ Name · test
✔ Make test an admin? · no
✔ Invite expires after · 14d
✔ Save peer invitation file to · test.toml
✔ Create peer test? · yes
adding to WireGuard interface: test (xxxxxx)

Peer "test" added
Peer invitation file written to test.toml
Please send it to them securely (eg. via magic-wormhole) to bootstrap them onto the network.

$ russ @ devmachine in ~ [18:29:59]
↳ sudo inn install test.toml             
[!] /etc/innernet is world-accessible (mode is 0o755). This is probably not what you want.
✔ Interface name · testinnernettest
[*] bringing up the interface.
[*] Generating new keypair.
[*] Registering keypair with server (at 10.42.0.1:51820).
[E] failed to start the interface: http://10.42.0.1:51820/v1/user/redeem: status code 401.
[*] bringing down the interface.
[E] Failed to redeem invite. Now's a good time to make sure the server is started and accessible!

[E] http://10.42.0.1:51820/v1/user/redeem: status code 401
```

Also thank you for this project! Rust, wireguard, peer to peer, it has everything!

mcginty:
Hey @russkel and sorry for the late reply! One can't run a client and server on the same machine (or, more specifically, the same network namespace) since you'll have conflicting routes (`ip route`), as both interfaces will be trying to route `10.42.0.0/16`.

While you could probably do some fancy stuff with [`ip-netns`](https://man7.org/linux/man-pages/man8/ip-netns.8.html), my general recommendation would be to have the innernet-server on a separate host.

russkel:
Thanks @mcginty. I see. Thanks for the reply. If I were to run the server in a docker container that will solve the namespace issues?

tommie:
The routes are already set up, right?

Isn't it enough to

1. let innernet-server also update the local `/etc/hosts`
2. change `SERVER_NAME` in `innernet-server new` to be customizable and default to the hostname?

[...]

PR #336, opened 2025-03-03
Periodically write a hostfile on the innernet server

bschwind:
Getting this started, spurred by a [comment](https://github.com/tonarino/innernet/issues/335#issuecomment-2694225949).

Related to #135 and #335

Right now this is implemented as a periodic tokio async task on the server when you run the `serve` command. I extracted the `update_hosts_file()` function from the client into the `shared` crate. That felt like the highest value location to add this functionality, though we could also add it to the various other server commands if we wish.

Beware, I haven't tested this yet at all.


server/src/db/peer.rs
```diff
@@ -338,6 +338,16 @@ impl DatabasePeer {
         Ok(peer_iter.collect::<Result<_, _>>()?)
     }
 
+    pub fn list_enabled(conn: &Connection) -> Result<Vec<Self>, ServerError> {
+        let mut stmt = conn.prepare_cached(&format!(
+            "SELECT {} FROM peers WHERE peers.is_redeemed = 1 AND peers.is_disabled = 0",
```
bschwind, line 343: Ideally we'd just do a `SELECT ip, name` but the deserialization machinery is all in place to select the whole peer so I just went with that for now for convenience. Performance likely wouldn't be a problem unless you have a _lot_ of peers.

alerque:
If there is some expectation that this is pulled together enough to maybe work I'll give it a test drive.

bschwind:
@alerque I would recommend testing it by creating a new network with this server binary, adding a peer, then running the serve command to see if it writes to the hostfile. If you don't get to it I'll test it out later.

alerque:
I have deployed e57ab94 on a production network by swapping out the system package for the last stable version (I am the Arch Linux packager for innernet) with an otherwise nearly identical package one built from this PR. I swapped out the package, restarted the server, and the `/etc/hosts` file and directs access to the innernet piers is working as expected! The catch of course is that things like `innernet list <network>` don't show anything at all and clients show it as `innernet-server.<network.wg` instead of what might be a preferred pier name. But having the current pier IP list in the hosts file is a *huge* improvement even without being a fully normalized member itself.

bschwind:
@alerque nice, thanks for testing! I'll see what I can do about the naming, but glad to hear it at least somewhat works.

strohel (approved):
Sorry for a late review - this looks good!

Initially I had a slight concern about 10 secs interval being a bit too low, but I see that we have an optimization that we don't write at all if the contents would be the same, so this boild to _reading_ /etc/hosts every 10 secs, which is very fine.

Ship it, especially if this has already received testing.

PR #337, opened 2025-03-11
publicip: use quad9 resolver instead of cloudflare

mcginty:
Quad9 is a not-for-profit resolver that is much less sketchy than Cloudflare, and I like their use of `A`/`AAAA` records better.

### DIFF
--- a/publicip/src/lib.rs
+++ b/publicip/src/lib.rs
@@ -6,9 +6,7 @@
 use std::{
     fs::File,
     io::{Cursor, Error, ErrorKind, Read, Write},
-    marker::PhantomData,
-    net::{AddrParseError, IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, UdpSocket},
-    str::FromStr,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, UdpSocket},
     time::Duration,
 };
 
@@ -20,12 +18,13 @@ macro_rules! ensure {
     };
 }
 
-const TYPE_TXT: u16 = 0x0010; // TXT-type requests (could also be A, AAAA, etc.)
-const CLASS_CH: u16 = 0x0003; // Because we are in the chaos realm.
+const CLASS_IN: u16 = 0x0001;
+const TYPE_A: u16 = 0x0001;
+const TYPE_AAAA: u16 = 0x001C;
 
-static CLOUDFLARE_QNAME: &[&str] = &["whoami", "cloudflare"];
-const CLOUDFLARE_IPV4: Ipv4Addr = Ipv4Addr::new(1, 1, 1, 1);
-const CLOUDFLARE_IPV6: Ipv6Addr = Ipv6Addr::new(0x2606, 0x4700, 0x4700, 0, 0, 0, 0, 0x1111);
+static QNAME: &[&str] = &["whatismyip", "on", "quad9", "net"];
+const IPV4_ADDRESS: Ipv4Addr = Ipv4Addr::new(9, 9, 9, 9);
+const IPV6_ADDRESS: Ipv6Addr = Ipv6Addr::new(0x2620, 0xfe, 0, 0, 0, 0, 0, 0xfe);
 
 pub enum Preference {
     Ipv4,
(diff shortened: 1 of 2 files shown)


publicip/src/lib.rs
```diff
@@ -137,24 +133,19 @@ impl<T: Into<IpAddr> + FromStr<Err = AddrParseError>> Request<T> {
         if qname_len & 0xc000 != 0xc000 {
             buf.set_position(buf.position() + qname_len as u64);
         }
-        ensure!(buf.read_u16()? == TYPE_TXT, "answer is not TXT type");
-        ensure!(buf.read_u16()? == CLASS_CH, "answer is not CH class");
+        ensure!(
+            buf.read_u16()? == self.record_type,
+            "answer is not expected type"
+        );
```
goodhoko, line 139: Would be nice to include the expected type in the error message like eg.

```suggestion
        let answer_type = buf.read_u16()?;
        ensure!(
            answer_type == self.record_type,
            "answer type ({answer_type}) is not expected type ({self.record_type})"
        );
```

but I see that would require touching up our `ensure!` macro which is probably out of scope here.

publicip/src/lib.rs
```diff
@@ -137,24 +133,19 @@ impl<T: Into<IpAddr> + FromStr<Err = AddrParseError>> Request<T> {
...
-        let end = start + txt_len;
-        ensure!(response.len() >= end, "unexpected txt answer lengths");
-
-        let txt = std::str::from_utf8(&response[start..end]).ok();
-        let answer = txt
-            .and_then(|txt| txt.parse::<T>().ok())
-            .ok_or_else(|| Error::new(ErrorKind::InvalidInput, "TXT not IP address".to_string()))?;
-
-        Ok(answer)
+        ensure!(data_len == N, "unexpected record data length");
```
goodhoko, line 146: same here

publicip/src/lib.rs
```diff
@@ -20,24 +18,25 @@ macro_rules! ensure {
...
+const CLASS_IN: u16 = 0x0001;
+const TYPE_A: u16 = 0x0001;
+const TYPE_AAAA: u16 = 0x001C;
 
-static CLOUDFLARE_QNAME: &[&str] = &["whoami", "cloudflare"];
-const CLOUDFLARE_IPV4: Ipv4Addr = Ipv4Addr::new(1, 1, 1, 1);
-const CLOUDFLARE_IPV6: Ipv6Addr = Ipv6Addr::new(0x2606, 0x4700, 0x4700, 0, 0, 0, 0, 0x1111);
+static QNAME: &[&str] = &["whatismyip", "on", "quad9", "net"];
+const IPV4_ADDRESS: Ipv4Addr = Ipv4Addr::new(9, 9, 9, 9);
+const IPV6_ADDRESS: Ipv6Addr = Ipv6Addr::new(0x2620, 0xfe, 0, 0, 0, 0, 0, 0xfe);
```
goodhoko, line 27:
```suggestion
// Reference: https://www.quad9.net/service/service-addresses-and-features
static QNAME: &[&str] = &["whatismyip", "on", "quad9", "net"];
const IPV4_ADDRESS: Ipv4Addr = Ipv4Addr::new(9, 9, 9, 9);
const IPV6_ADDRESS: Ipv6Addr = Ipv6Addr::new(0x2620, 0xfe, 0, 0, 0, 0, 0, 0xfe);
```

publicip/src/lib.rs
goodhoko, line 1: GitHub won't let me comment on these lines, but this comment at the very top of the file should be updated too:

```
//! Get your public IP address(es) as fast as possible, with no dependencies.
//!
//! Currently uses Cloudflare's DNS as it's the simplest, but that could change
//! in the future.
```

goodhoko, line 1: There's also this prompt that should be updated:
https://github.com/tonarino/innernet/blob/b2e414984e007121ce8b7f84877bfe935d209c17/shared/src/prompts.rs#L564

goodhoko (commented):
The intention and code both look good to me, thanks!

The practical implications of this for innernet are not entirely clear to me. It looks like `publicip` is only used during server initialization and when overriding endpoint which are both user-driven actions, so this shouldn't affect the "normal" service of innernet. We should definitely test both of these use-cases before merging.

bschwind:
@mcginty you should be able to rebase on the latest main to clear up the CI issues.

mcginty:
@goodhoko Yep it's only used for initializing an innernet network, I initially wrote it as a challenge to myself when I was trying to keep innernet's dependencies to an absolute minimum, so `publicip`'s main purpose is to be a fast and zero-dependency external IP resolver.

I switched to Quad9 here because I was talking to them at a conference and like them better than CloudFlare, so I took it as a small shoutout to them to implement this and then I was thinking of releasing it as a separate crate for others to use.

publicip/src/lib.rs
mcginty: yep that was my rationale for the ambiguous message too :P.

mcginty:
Updated, sorry for the delay.

goodhoko (approved):
All my remarks were addressed. I'd like others to chime in for the final decision. 🙏 @strohel @bschwind

PR #330, opened 2024-12-16
Fix Dockerfile for arm64 compatibility

eos175:

PR #340, opened 2025-03-13
Fix Dockerfile for arm64 compatibility

bschwind:
Originally from #330


strohel (approved):
LGTM again.

Does GitHub CI support ARM64 runners? I would be nice to also test this on CI, but of course can be done later.

PR #342, opened 2025-03-26
Add cli to publicip crate

Techcable:
Although not the officially intended purpose, this makes a nifty little tool :)


bschwind (approved):
LGTM, nice!

PR #352, opened 2025-08-06
Restore state from cache when the server is not reachable

### LINKED ISSUE #296: Attempt NAT traversal when fetching, even if server is inaccessible
Scenario: You have machines all in one physical building on a LAN, all connected via innernet. Maybe each peer connects to each other peer's _external_ endpoint. innernet-server runs in a separate network.

If the building's external internet connection goes down, but machines are still connected via the LAN, the peers within that LAN should all be able to talk to each other by running NAT discovery with endpoints they've cached from the innernet server. Currently peers will just lose connectivity to each other, because although they have the _internal_ endpoint cached in `/var/lib/innernet/network_name.json`, they won't do anything with it if the connection to the server fails.

I think we just need to modify the fetch functionality [here](https://github.com/tonarino/innernet/blob/489f505f67411f179eff147c2439437bd65aa418/client/src/main.rs#L570-L571) to run client NAT traversal in case the innernet server can't be contacted. Otherwise, the client `fetch` command fails and causes the binary to exit without attempting to reconnect to peers.

### LINKED ISSUE #316: Restore state from cache on startup when server is unreachable
Innernet client is unable to setup the network interface when it can't reach innernet server on startup. This is a single-point of failure even when other peers are perfectly reachable.

At the same time innernet already caches peer information fetched from the server in `/var/lib`. We could use this cache on startup when we fail to read state from server.

cc @skywhale @strohel @bschwind

mbernat:
Closes #316 ...and also closes #296, I guess.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -588,7 +588,7 @@ fn fetch(
             if store.peers().is_empty() {
                 bail!(
                     "Could not connect to the innernet server and there are no cached peers, \
-                     bailing."
+                     exiting."
                 )
             }
 
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -582,7 +582,20 @@ fn fetch(
     );
     let mut store = DataStore::open_or_create(&opts.data_dir, interface)?;
     let api = Api::new(&config.server);
-    let State { peers, cidrs } = api.http("GET", "/user/state")?;
+    let (State { peers, cidrs }, server_is_reachable) = match api.http("GET", "/user/state") {
+        Ok(state) => (state, true),
+        Err(ureq::Error::Transport(_)) => {
+            log::warn!(
+                "Could not connect to the innernet server, proceeding with cached state instead."
+            );
+            let state = State {
+                peers: store.peers().to_vec(),
+                cidrs: store.cidrs().to_vec(),
+            };
+            (state, false)
+        },
+        Err(e) => bail!(e),
+    };
 
     let device = Device::get(interface, opts.network.backend)?;
     let modifications = device.diff(&peers);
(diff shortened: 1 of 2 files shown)


client/src/main.rs
```diff
@@ -582,7 +582,20 @@ fn fetch(
...
+    let (State { peers, cidrs }, server_is_reachable) = match api.http("GET", "/user/state") {
+        Ok(state) => (state, true),
+        Err(ureq::Error::Transport(_)) => {
+            log::warn!(
+                "Could not connect to the innernet server, proceeding with cached state instead."
+            );
+            let state = State {
+                peers: store.peers().to_vec(),
+                cidrs: store.cidrs().to_vec(),
+            };
```
strohel, line 594: What happens if we don't have any cache here? I guess that will still bring the interface up, just report 0 peers? Do we want to explicitly bail instead?

strohel (commented):
Oh wow, is it _that_ easy?! Niice.

Adding one question. Will be happy to test this together on Friday.

bschwind (approved):
LGTM, nice!

client/src/main.rs
mbernat: This will be the new behavior.

```
marek@Mac innernet % cargo build && sudo ./target/debug/innernet up tonari  
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
[*] fetching state for tonari from server...

[E] Could not connect to the innernet server and there are no cached peers, bailing.
```

The interface itself is brought up first and keep up even if peer initialization fails and I guess that's intentional so that the subsequent fetches (which don't don't support bringing interface up automatically, for whatever reason) have a chance to succeed?

client/src/main.rs
```diff
@@ -585,6 +585,13 @@ fn fetch(
     let (State { peers, cidrs }, server_is_reachable) = match api.http("GET", "/user/state") {
         Ok(state) => (state, true),
         Err(ureq::Error::Transport(_)) => {
+            if store.peers().is_empty() {
+                bail!(
+                    "Could not connect to the innernet server and there are no cached peers, \
+                     bailing."
```
strohel, line 591: I think the idiom is
```suggestion
                     bailing out."
```
...i.e. that it has a different meaning without the "out"?

client/src/main.rs
strohel: New behavior looks good!

> The interface itself is brought up first and kept up even if peer initialization fails

Not sure if that was the _original_ intention, but now such a behavior is beneficial and depended on: you can start a program that binds to the listen port etc. So let's not change it here.

strohel (approved):
Nice!

client/src/main.rs
mbernat: This app seems to use "exiting" in many other places, so I'll stick with that.

PR #353, opened 2025-08-07
Add a /user/capabilities endpoint with server capabilities

mbernat:
An alternative to #351, on top of #315.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -1027,28 +1027,28 @@ fn prompt_override_endpoint(
     args: &OverrideEndpointOpts,
     listen_port: u16,
 ) -> Result<Option<Endpoint>, Error> {
+    let unspecified_ip_supported = server_capabilities.unspecified_ip_in_override_endpoint;
+
     let endpoint = match &args.endpoint {
         Some(endpoint) => endpoint.clone(),
         None => {
-            let unspecified_ip_supported = server_capabilities.unspecified_ip_in_override_endpoint;
-
             let external_ip = if unspecified_ip_supported {
                 prompts::unspecified_ip_and_auto_detection_flow()?
             } else {
                 prompts::ip_auto_detection_flow()?
             };
 
-            let endpoint = prompts::input_external_endpoint(external_ip, listen_port)?;
-            if endpoint.is_host_unspecified() && !unspecified_ip_supported {
-                bail!(
-                    "Attempted to use an unspecified IP (all zeros) but the innernet server does
-    not have the capability to resolve it, likely because its version is older."
-                )
-            }
-
-            endpoint
+            prompts::input_external_endpoint(external_ip, listen_port)?
         },
     };
+
+    if endpoint.is_host_unspecified() && !unspecified_ip_supported {
+        bail!(
+            "Attempted to use an unspecified IP (all zeros) but the innernet server does
+    not have the capability to resolve it, likely because its version is older."
+        )
+    }
+
     if args.yes || prompts::confirm(&format!("Set external endpoint to {endpoint}?"))? {
         Ok(Some(endpoint))
     } else {


client/src/main.rs
```diff
@@ -1035,6 +1037,52 @@ fn override_endpoint(
     Ok(())
 }
 
+fn prompt_override_endpoint(
+    info: &Info,
```
strohel, line 1041: Super nit: `server_info` for the variable name? Or maybe even for the type name if you feel like it.

client/src/main.rs
```diff
@@ -1009,15 +1009,17 @@ fn override_endpoint(
     sub_opts: OverrideEndpointOpts,
 ) -> Result<(), Error> {
     let config = InterfaceConfig::from_interface(&opts.config_dir, interface)?;
+    // TODO(mbernat): Refactor command handling so that we can gather the server info in `run()`.
+    let info = get_server_info(&config)?;
 
     let endpoint_contents = if sub_opts.unset {
-        prompts::unset_override_endpoint(&sub_opts)?.then_some(EndpointContents::Unset)
+        prompt_unset_override_endpoint(&sub_opts)?.then_some(EndpointContents::Unset)
     } else {
```
strohel, line 1017: Nit: we don't have to call the /info endpoint when unsetting override.

shared/src/types.rs
```diff
@@ -765,6 +765,12 @@ pub struct State {
     pub cidrs: Vec<Cidr>,
 }
 
+/// This model is sent as a response to the /info endpoint.
+#[derive(Debug, Default, Clone, Deserialize, Serialize)]
+pub struct Info {
+    pub override_endpoint_with_unspecified_ip_is_supported: bool,
```
strohel, line 771: Nit: add `#[serde(default)]` to cope with the very hypothetical situation the field is removed (false is a good default).

client/src/main.rs
```diff
@@ -1035,6 +1037,52 @@ fn override_endpoint(
...
+
+fn get_server_info(config: &InterfaceConfig) -> Result<Info, Error> {
+    let api = Api::new(&config.server);
+    let maybe_info: Result<Info, ureq::Error> = api.http("GET", "/user/info");
+    match maybe_info {
+        Ok(info) => Ok(info),
+        Err(ureq::Error::Status(404, _)) => {
+            log::warn!(
+                "innernet server info is missing, the server might not support all the client features"
+            );
```
strohel, line 1077: Maybe this should be just a debug, on its own it isn't very informative.

client/src/main.rs
```diff
@@ -1035,6 +1037,52 @@ fn override_endpoint(
     Ok(())
 }
 
+fn prompt_override_endpoint(
+    info: &Info,
+    args: &OverrideEndpointOpts,
+    listen_port: u16,
+) -> Result<Option<Endpoint>, Error> {
+    let endpoint = match &args.endpoint {
+        Some(endpoint) => endpoint.clone(),
```
strohel, line 1046: I guess we can/should bail out if an unspecified addr is set _on the command line_ and the server doesn't support that.

strohel (commented):
Yeah I think I prefer this variant! The advantage is that this allows us to _remove_ capabilities in new server and this will still be seen by older clients. (i.e. it it let's the server be the source of thruth, which is more manageable)

Just come small suggestions.

client/src/main.rs
strohel: @mbernat I meant when someone types: `innernet override-endpoint --endpoint 0.0.0.0:1234` -- I think we don't catch this case currently and let it fall through to the (old) server.

mbernat:
### Talking to an older server

```
marek@Mareks-Laptop innernet % cargo build -q && sudo ./target/debug/innernet -v override-endpoint tabletop
[D] innernet server endpoint capabilities not found, assuming default capabilities
? Auto-detect external endpoint IP address (via DNS query to 9.9.9.9)? (y/n) ›
```

### Talking to a newer server

```
marek@Mareks-Laptop innernet % cargo build -q && sudo ./target/debug/innernet -v override-endpoint tabletop
[*] Note: use this option if you do not have a fixed global IP but the port is forwarded.
? Use an unspecified IP address and override just the port? (y/n) ›
```

client/src/main.rs
mbernat:
```
marek@Mareks-Laptop innernet % cargo build -q && sudo ./target/debug/innernet -v override-endpoint --endpoint 0.0.0.0:1 tabletop
[D] innernet server endpoint capabilities not found, assuming default capabilities

[E] Attempted to use an unspecified IP (all zeros) but the innernet server does
    not have the capability to resolve it, likely because its version is older.
```

strohel (approved):
This is perfect!

I'll let you merge on your discretion w.r.t. parent PRs/rebasing/testing.

Issue #338, opened 2025-03-13
The override-endpoint command seems to get stuck with console >= 0.15.8


bschwind:
I was testing #337 and tried running the `override-endpoint` command like so:

```
brian innernet $ sudo innernet override-endpoint tonari
✔ Auto-detect external endpoint IP address (via a DNS query to 1.1.1.1)? · yes
? External endpoint › 10.20.30.40:51820
```

(IP intentionally changed)

When I pressed enter to confirm the external endpoint, the program just got stuck and no longer accepted input, but also didn't time out.

mbernat:
Further tracked to this PR's merge commit: https://github.com/console-rs/console/pull/194

Does not make a ton of sense: the PR only updates the MSRV, windows deps and some cosmetic changes to one unix file :/

mbernat:
> An interim work-around would be to explicitly depend on console < 0.15.8. Should I fire a PR for that?

Go for it!

mbernat:
This is now technically fixed by a workaround but keeping the issue open until I report it upstream.

PR #359, opened 2025-08-12
Use interact_text() to work around an initial text bug

mbernat:
- An alternative fix for #338 
- Also allows using arrow keys for editing the initial input
- Docs mention that `interact_text()` limits text inputs to ASCII, which might be undesirable, but it does not seem to be true in practice

> Its difference from [interact](https://docs.rs/dialoguer/latest/dialoguer/struct.Input.html#method.interact) is that it only allows ascii characters for string, while [interact](https://docs.rs/dialoguer/latest/dialoguer/struct.Input.html#method.interact) allows virtually any character to be used e.g arrow keys.

### DIFF
--- a/server/src/initialize.rs
+++ b/server/src/initialize.rs
@@ -146,6 +146,7 @@ pub fn init_wizard(conf: &ServerConfig, opts: InitializeOpts) -> Result<(), Erro
         Input::with_theme(&theme)
             .with_prompt("Network CIDR")
             .with_initial_text("10.42.0.0/16")
+            // TODO(mbernat): workaround for https://github.com/console-rs/dialoguer/issues/330
             .interact_text()?
     };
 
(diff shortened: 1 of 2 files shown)


server/src/initialize.rs
```diff
@@ -144,7 +144,7 @@ pub fn init_wizard(conf: &ServerConfig, opts: InitializeOpts) -> Result<(), Erro
         Input::with_theme(&theme)
             .with_prompt("Network CIDR")
             .with_initial_text("10.42.0.0/16")
-            .interact()?
+            .interact_text()?
```
strohel, line 147: Optionally we could add a TODO here and below mentioning we work-around the issue (and link it), mention it limits to ASCII.

mbernat:
> How does this behave when you try to enter something non-ASCII? It just ignores it?

It seems to accept it just fine, actually, at least when I use macos emoji input or copy&paste something. Maybe the docs are out of date.

Moreover, I just discovered that while the fix in 1.7.0 allows you to accept the initial input, it doesn't allow you to use arrow keys to modify the middle of the initial input, whereas this PR fixes that too.

PR #365, opened 2025-08-23
Derive Hash for wireguard_control::Key

alpn:


strohel (approved):
Yeah makes sense. I wouldn't oppose also deriving `PartialOrd, Ord`, but that can as well be done separetely.

PR #378, opened 2026-01-07
Fix bug that prevented preserving SELinux context and add tests

jebotz:
Fixed a small bug in `get_temp_path()` where temp files were created in the wrong directory. This caused `hosts_dir.with_file_name(temp_filename)` to produce `/hosts.tmpXXX` instead of `/etc/hosts.tmpXXX`, which in turn broke SELinux context preservation.

Also added tests to verify temp file location and SELinux context preservation.


strohel (approved):
Oh great catch, thanks (also for the tests)!

I'll wait some time if others also want to review, and merge afterwards unless if nothing pops up.

strohel:
Let me close this due to inactivity, but @jebotz if/once you have free cycles for this, we'll happily reopen.

Issue #334, opened 2025-02-22
Unhelpful 400 error when trying to add-cidr, unclear behavior


jakintosh:
Hi all,

First—I love this project. It's rare to find something that both feels like magic and is also so simple to see exactly how and why it works. So glad to have an option like this over something like Tailscale.

Anyhow, I'm in the early stages of deploying/figuring out how I want to partition my innernet's CIDRs, and am running into an issue where I'm getting a 400 from the server when trying to add a new CIDR.

## Current Network

```
mynetwork (:xxx)
  10.27.0.0/16 mynetwork
    10.27.0.1/32 innernet-server
    | ◉ 10.27.0.1: innernet-server (xxx…)
    10.27.8.0/24 fleet
    | ◉ 10.27.8.1: a (you, xxx…)
    | ◉ 10.27.8.2: b (xxx…)
    10.27.16.0/24 cloud
    | ◉ 10.27.16.1: c (xxx…)
```

## Goal

What I want to do next, after I started adding machines, is to add a layer of granularity to my `cloud` block, one for my physical servers that I can see and touch, and another for my virtual servers that I rent from the public cloud. What I'd really like to do is add two CIDRs to 'cloud', a 'physical' 10.27.16.0/25 and a 'virtual' 10.27.16.128/25 block. However, I couldn't figure out if there was a way to "move" an existing peer (in this case 'c') into a child CIDR, even if its IP doesn't change.

## Issue

When I try to run either:

`sudo innernet add-cidr mynetwork --name physical --cidr '10.27.16.0/25' --parent 'cloud' --yes`

or

`sudo innernet add-cidr mynetwork --name virtual --cidr '10.27.16.128/25' --parent 'cloud' --yes`

I get

`[E] http://10.27.0.1:51820/v1/admin/cidrs: status code 400`

I feel like I'm getting bitten by the immutability of the network, even though I'm not technically trying to change any IP addresses.

Is my only option to nuke my config and start over? Presumably, with a little more future planning this time.

Thanks!

strohel:
Agreed the errors are often unhelpful: the server should propagate them better to the client (patches welcome!). Another example of unhelpful is when adding a peer with a name that already exists (even if it is disabled).

About your particular issue: I think it is how you describe it: innernet currently doesn't support moving a peer between CIDRs (not even special cases where its IP wouldn't change - AFAIK). In these situations we disable old peer and create a new peer (renaming the original to something like `peername-old` first to free up the name).

[...]

PR #389, opened 2026-03-10
Refactor `create_peer` API

mbernat:
- Follows the remaining suggestions in https://github.com/tonarino/innernet/pull/382#discussion_r2859409122
- Adds a custom error type
- Also refactors a bunch of internal methods

## Testing

Check error paths

- [x] IPv6 address on IPv4 innernet network
- [x] IPv4 address on IPv6 innernet network
- [x] IP address that doesn't belong to the provided CIDR
- [x] Non-existing CIDR ID

### DIFF
--- a/client-core/examples/add_peer.rs
+++ b/client-core/examples/add_peer.rs
@@ -2,9 +2,9 @@ use anyhow::{Context, Error};
 use env_logger::Env;
 use innernet_client_core::{
     interface::{InterfaceConfig, InterfaceName},
-    peer::{self, NewPeerInfo},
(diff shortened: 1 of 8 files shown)


strohel (commented):
Looks good in general! Some technicalities.

strohel (approved):
LGTM! Ship it!

Thanks for also testing requests with invalid payloads, it looks good! Ideally the server also returns text with the error description, but that's orthogonal and already partly tracked in https://github.com/tonarino/innernet/issues/334

PR #393, opened 2026-04-18
Ensure host architecture == target arch in RPM build script

eos175:


strohel (approved):
Sounds sensible, though I have very little context about the RPM build scripts.

BTW where does limitation of not supporting cross compilation come from? From using `docker build`?

FWIW I'd be also very open to move the RPM build infrastructure to an external git repo, where someone more versed with RPM-based distros could maintain it. This seems to work well for [Deb packages](https://github.com/tommie/innernet-debian).

strohel:
> FWIW I'd be also very open to move the RPM build infrastructure to an external git repo, where someone more versed with RPM-based distros could maintain it. This seems to work well for [Deb packages](https://github.com/tommie/innernet-debian).

This is related to https://github.com/tonarino/innernet/issues/203, I see @wwalker and @refi64 were active to an extent on RPM builds.

eos175:
Hi @strohel,

The main reason is that cross-compiling can be quite slow and often creates compatibility issues across different distributions. For me, managing all those variations is complicated and difficult to maintain.

I believe the most reliable approach for now is to simply build the package on the specific platform and architecture where it's needed. This check just ensures the script runs in the correct environment to avoid broken builds.

I'm not planning on making more complex changes, so this small validation helps keep things stable for anyone using the script as it is.

strohel:
Alright, let me merge this then, sorry for delays.

PR #394, opened 2026-05-18
Fix reverse path filtering documentation in readme

strohel:
- relates to https://github.com/tonarino/innernet/issues/26
- sibling to (tonari internal PR): https://github.com/tonarino/portal/pull/6166

By chance I've found out it is not sufficient to set `rp_filter` just for `all` and `default`, that doesn't apply to _existing_ devices. Update the advice.

I've also checked and `*` does **not** apply to `all` and `default`, so keep these.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,8 +168,10 @@ If you're running a service on innernet, there are some important security consi
 Strict RPF prevents packets from _other_ interfaces from having internal source IP addresses. This is _not_ the default on Linux, even though it is the right choice for 99.99% of situations. You can enable it by adding the following to a `/etc/sysctl.d/60-network-security.conf`:
 
 ```
-net.ipv4.conf.all.rp_filter=1
-net.ipv4.conf.default.rp_filter=1
+net.ipv4.conf.all.rp_filter = 1
+net.ipv4.conf.default.rp_filter = 1
+# also apply to existing interfaces:
+net.ipv4.conf.*.rp_filter = 1
 ```
 
 ### Bind to the WireGuard device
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -168,8 +168,10 @@ If you're running a service on innernet, there are some important security consi
 Strict RPF prevents packets from _other_ interfaces from having internal source IP addresses. This is _not_ the default on Linux, even though it is the right choice for 99.99% of situations. You can enable it by adding the following to a `/etc/sysctl.d/60-network-security.conf`:
 
 ```
-net.ipv4.conf.all.rp_filter=1
-net.ipv4.conf.default.rp_filter=1
+net.ipv4.conf.all.rp_filter = 1
+net.ipv4.conf.default.rp_filter = 1
+# also apply to existing interfaces:
```
mbernat: This comment is rather cryptic. What does 'existing' mean precisely and why would you want to apply the other rules to a non-existent interface? Is that for when you plug an interface in at runtime?

We might also want to mention that `all/default` and `*` are exactly complementary.

mbernat (approved):
LGTM

README.md
strohel: Like this?
```suggestion

# Also apply to interfaces that exist at the time this sysctl is applied, as they might have been
# created with value 2 (loose), which takes precedence even when all.rp_filter is 1.
# Note that using the * match does not apply to the special `all` and `default` knobs.
```

bschwind (approved):
LGTM!

Issue #314, opened 2024-05-30
Feature: Assigning external IP addresses (or address ranges) to a peer


agausmann:
Motivating example: I have an external network, and I want to make the external network and the innernet network fully routable between each other. 

The idea I have is to use one innernet peer as a relay/router that forwards traffic between the two networks. I can set up the routes on my external network just fine, but the innernet peers also need to be configured to send traffic for the other network to the relay peer. 

(For simplicity, let's assume the external network is on a separate subnet, not a part of the innernet subnet, and the only machine that is a member of both networks is the relay.)

To implement this, the additional addresses for the peer could be registered in the innernet server database, and then distributed to the other peers as "additional routes" of the relay peer. The innernet clients can add those addresses to the "Allowed IPs" list of the relay peer and add the corresponding routes in the OS network interface.

This is related to #2, #3, and a little bit of #210. If that external network I described before is a "vanilla Wireguard" network, then this can help bring vanilla Wireguard clients (including Android) onto the network. By using a relay server, the vanilla configuration stays simple and static, and doesn't require manual work or a daemon to maintain the full list of peers and do the whole NAT traversal dance.

Other alternatives: It is probably possible to use NAT on the relay, so all the traffic on the innernet side goes through the relay's primary innernet address. That has the usual drawbacks of NAT: only the external network can establish a connection to the innernet network, and not vice versa (besides port forwarding).

agausmann:
This is not an important feature for me right now, but the problem of android and vanilla connections has been in the back of my mind for a while, and I had this idea that I wanted to share.

If there's other people that would be interested in it, then I wouldn't mind helping out. In the meantime, I may put together a repo with some scripts/configs for a relay server based on NAT.

PR #395, opened 2026-05-29
Client override endpoint

bschwind:
Related to #314

This lets an innernet client specify a particular endpoint to use for a peer.

Motivating example: When running a relay like [wpex](https://github.com/weiiwang01/wpex), you must specify the _relay endpoint_ as the endpoint for the peer you want to talk to over the relay, and they must do the same for your peer.

To set an override:

```
$ sudo innernet override-peer-endpoint <INTERFACE_NAME>
```

`name` and `endpoint` can also be specified, along with a `--yes` flag to bypass confirmation:

```
$ sudo innernet override-peer-endpoint <INTERFACE_NAME> --name <PEER_NAME> --endpoint <ENDPOINT> --yes
```

To unset an override:

```
$ sudo innernet override-peer-endpoint <INTERFACE_NAME> --unset
```

When unsetting, you can also use `--name` and the `--yes` flag.

```
$ sudo innernet override-peer-endpoint <INTERFACE_NAME> --unset --name <PEER_NAME> --yes
```

I ultimately decided to implement this by adding a simple `HashMap<IpAddr, Endpoint>` to the `InterfaceConfig`. At first I wanted something stored on `Peer`, but that type is shared with the server and everything else, where a "local peer override" doesn't really make sense. The server should not be involved or care at all if a client overrides a particular peer's endpoint.

When we override an endpoint, we immediately tell the wireguard device to set that endpoint for the peer. When `fetch`ing from the innernet server, we apply the overrides from the `InterfaceConfig` on the list of peers before showing a diff or doing any NAT traversal.

~~One open question: if the specified endpoint is not reachable, should we allow NAT traversal to run and potentially set the endpoint to something different which _is_ reachable? Or should it remain fixed on that specified endpoint override no matter what?~~

We've decided to skip NAT traversal if a given peer has an endpoint override in place.

## Testing
- [x] does `override-peer-endpoint` work when the respective interface is down?
- [x] try to unset a peer endpoint override that isn't set in the first, place specifycing `--name` on the command line. Return value should be 0.
- [x] NAT traversal is indeed skipped for peers with locally overriden endpoint (try with peer that is offline but has multiple candidates)
- [x] for peers A and B who cannot reach each other, their override-peer-endpoint to a `wpex` instance at both sides, they should connect immediately (even with no background `innernet fetch` / daemon)
  - [x] reboot peer A. The connection through wpex should resume after booting
  - [x] reboot both peers simultaneously, connection should resume
- [ ] for peers A and B that already have connection between each other, try to use `override-peer-endpoint` on both sides to route their traffic through wpex instance. Describe what happens. (they may "stick" to their existing endpoints)
  - [ ] if the endpoint was indeed sticky above, try `innernet down <interface>` first at both ends, then override peer endpoints

### DIFF
--- a/client-core/src/data_store.rs
+++ b/client-core/src/data_store.rs
@@ -19,7 +19,7 @@ struct FileBackedData<T> {
     contents: T,
 }
 
-impl<T: Default + DeserializeOwned> FileBackedData<T> {
+impl<T: Default + DeserializeOwned + Serialize> FileBackedData<T> {
     fn open_with_path<P: AsRef<Path>>(path: P, create: bool) -> Result<Self, WrappedIoError> {
         let path = path.as_ref();
         let is_existing_file = path.exists();
@@ -42,6 +42,14 @@ impl<T: Default + DeserializeOwned> FileBackedData<T> {
         let contents = serde_json::from_str(&json).unwrap_or_else(|_| T::default());
 
         Ok(Self { file, contents })
+    }
+
+    fn write(&mut self) -> Result<(), io::Error> {
+        self.file.rewind()?;
+        self.file.set_len(0)?;
+        self.file
+            .write_all(serde_json::to_string_pretty(&self.contents)?.as_bytes())?;
+        Ok(())
     }
 }
 
@@ -102,12 +110,11 @@ impl DataStore {
 
     /// Produce the path "`data_dir`/`interface`.peer-endpoint-overrides.json".
     pub fn get_peer_endpoint_overrides_path(data_dir: &Path, interface: &InterfaceName) -> PathBuf {
-        let file_name = format!("{interface}.peer-endpoint-overrides");
-        data_dir.join(file_name).with_extension("json")
-    }
+        let file_name = format!("{interface}.peer-endpoint-overrides.json");
 
-    pub fn peers_and_cidrs_file_exists(data_dir: &Path, interface: &InterfaceName) -> bool {
-        Self::get_peers_and_cidrs_path(data_dir, interface).exists()
+        // Warning - if you use `.with_extension("json")` here instead of inlining it above,
+        //           it will result in `{interface}.json` which is undesirable.
+        data_dir.join(file_name)
     }
 
     fn _open(
@@ -116,13 +123,16 @@ impl DataStore {
         create: bool,
     ) -> Result<Self, WrappedIoError> {
         ensure_dirs_exist(&[data_dir])?;
+
         let peers_and_cidrs = FileBackedData::open_with_path(
             Self::get_peers_and_cidrs_path(data_dir, interface),
             create,
         )?;
+
         let peer_endpoint_overrides = FileBackedData::open_with_path(
             Self::get_peer_endpoint_overrides_path(data_dir, interface),
-            create,
+            // Always create the peer overrides file if it doesn't exist.
+            true,
         )?;
 
         Ok(Self {
@@ -131,12 +141,12 @@ impl DataStore {
         })
     }
 
-    /// Open the data store residing at "`data_dir`/`interface`.json".
+    /// Open the data store residing in the "`data_dir`" directory.
     pub fn open(data_dir: &Path, interface: &InterfaceName) -> Result<Self, WrappedIoError> {
         Self::_open(data_dir, interface, false)
     }
 
-    /// Open or create the data store residing at "`data_dir`/`interface`.json".
+    /// Open or create the data store residing in the "`data_dir`" directory.
     pub fn open_or_create(
         data_dir: &Path,
         interface: &InterfaceName,
@@ -212,11 +222,7 @@ impl DataStore {
     }
 
     fn write_peers_and_cidrs(&mut self) -> Result<(), io::Error> {
-        self.peers_and_cidrs.file.rewind()?;
-        self.peers_and_cidrs.file.set_len(0)?;
-        self.peers_and_cidrs
-            .file
-            .write_all(serde_json::to_string_pretty(&self.peers_and_cidrs.contents)?.as_bytes())?;
+        self.peers_and_cidrs.write()?;
         Ok(())
     }
 
@@ -256,11 +262,7 @@ impl DataStore {
     }
 
     pub fn write_peer_endpoint_overrides(&mut self) -> Result<(), io::Error> {
-        self.peer_endpoint_overrides.file.rewind()?;
-        self.peer_endpoint_overrides.file.set_len(0)?;
-        self.peer_endpoint_overrides.file.write_all(
-            serde_json::to_string_pretty(&self.peer_endpoint_overrides.contents)?.as_bytes(),
-        )?;
+        self.peer_endpoint_overrides.write()?;
         Ok(())
     }
 
--- a/client-core/src/data_store.rs
+++ b/client-core/src/data_store.rs
@@ -159,6 +159,15 @@ impl DataStore {
         }
     }
 
+    pub fn endpoint_override_for_peer(&self, peer_ip: IpAddr) -> Option<SocketAddr> {
+        match &self.contents {
+            Contents::V1 {
+                local_endpoint_overrides,
+                ..
+            } => local_endpoint_overrides.get(&peer_ip).copied(),
+        }
+    }
+
     fn set_cidrs(&mut self, new_cidrs: Vec<Cidr>) {
         match &mut self.contents {
             Contents::V1 { ref mut cidrs, .. } => *cidrs = new_cidrs,
--- a/client-core/src/data_store.rs
+++ b/client-core/src/data_store.rs
@@ -198,7 +198,7 @@ impl DataStore {
         }
     }
 
-    fn write(&mut self) -> Result<(), io::Error> {
+    pub fn write(&mut self) -> Result<(), io::Error> {
         self.file.rewind()?;
         self.file.set_len(0)?;
         self.file
--- a/client-core/src/data_store.rs
+++ b/client-core/src/data_store.rs
@@ -1,11 +1,13 @@
 use anyhow::{bail, Error};
-use innernet_shared::{chmod, ensure_dirs_exist, Cidr, IoErrorContext, Peer, WrappedIoError};
+use innernet_shared::{
+    chmod, ensure_dirs_exist, Cidr, Endpoint, IoErrorContext, Peer, WrappedIoError,
+};
 use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
     fs::{File, OpenOptions},
     io::{self, Read, Seek, Write},
-    net::{IpAddr, SocketAddr},
+    net::IpAddr,
     path::{Path, PathBuf},
 };
 use wireguard_control::InterfaceName;
@@ -27,7 +29,7 @@ enum Contents {
         peers: Vec<Peer>,
         cidrs: Vec<Cidr>,
         #[serde(default)]
-        local_endpoint_overrides: HashMap<IpAddr, SocketAddr>,
+        local_endpoint_overrides: HashMap<IpAddr, Endpoint>,
     },
 }
 
@@ -148,23 +150,45 @@ impl DataStore {
         }
     }
 
-    pub fn local_endpoint_overrides(&self) -> impl Iterator<Item = (IpAddr, SocketAddr)> + '_ {
+    pub fn local_endpoint_overrides(&self) -> impl Iterator<Item = (IpAddr, Endpoint)> + '_ {
         match &self.contents {
             Contents::V1 {
                 local_endpoint_overrides,
                 ..
             } => local_endpoint_overrides
                 .iter()
-                .map(|(ip, socket)| (*ip, *socket)),
+                .map(|(ip, socket)| (*ip, socket.clone())),
         }
     }
 
-    pub fn endpoint_override_for_peer(&self, peer_ip: IpAddr) -> Option<SocketAddr> {
+    pub fn endpoint_override_for_peer(&self, peer_ip: IpAddr) -> Option<Endpoint> {
         match &self.contents {
             Contents::V1 {
                 local_endpoint_overrides,
                 ..
-            } => local_endpoint_overrides.get(&peer_ip).copied(),
+            } => local_endpoint_overrides.get(&peer_ip).cloned(),
+        }
+    }
+
+    pub fn override_endpoint_for_peer(&mut self, peer_ip: IpAddr, endpoint: Endpoint) {
+        match &mut self.contents {
+            Contents::V1 {
+                local_endpoint_overrides,
+                ..
+            } => {
+                local_endpoint_overrides.insert(peer_ip, endpoint);
+            },
+        }
+    }
+
+    pub fn unset_endpoint_for_peer(&mut self, peer_ip: IpAddr) {
+        match &mut self.contents {
+            Contents::V1 {
+                local_endpoint_overrides,
+                ..
+            } => {
+                local_endpoint_overrides.remove(&peer_ip);
+            },
         }
     }
 
--- a/client-core/src/data_store.rs
+++ b/client-core/src/data_store.rs
@@ -2,8 +2,10 @@ use anyhow::{bail, Error};
 use innernet_shared::{chmod, ensure_dirs_exist, Cidr, IoErrorContext, Peer, WrappedIoError};
 use serde::{Deserialize, Serialize};
 use std::{
+    collections::HashMap,
     fs::{File, OpenOptions},
     io::{self, Read, Seek, Write},
+    net::{IpAddr, SocketAddr},
     path::{Path, PathBuf},
 };
 use wireguard_control::InterfaceName;
@@ -21,7 +23,12 @@ pub struct DataStore {
 #[serde(tag = "version")]
 enum Contents {
     #[serde(rename = "1")]
-    V1 { peers: Vec<Peer>, cidrs: Vec<Cidr> },
+    V1 {
+        peers: Vec<Peer>,
+        cidrs: Vec<Cidr>,
+        #[serde(default)]
+        local_endpoint_overrides: HashMap<IpAddr, SocketAddr>,
+    },
 }
 
 impl DataStore {
@@ -47,6 +54,7 @@ impl DataStore {
         let contents = serde_json::from_str(&json).unwrap_or_else(|_| Contents::V1 {
             peers: vec![],
             cidrs: vec![],
+            local_endpoint_overrides: HashMap::new(),
         });
 
         Ok(Self { file, contents })
@@ -137,6 +145,17 @@ impl DataStore {
     pub fn cidrs(&self) -> &[Cidr] {
         match &self.contents {
             Contents::V1 { cidrs, .. } => cidrs,
+        }
+    }
+
+    pub fn local_endpoint_overrides(&self) -> impl Iterator<Item = (IpAddr, SocketAddr)> + '_ {
+        match &self.contents {
+            Contents::V1 {
+                local_endpoint_overrides,
+                ..
+            } => local_endpoint_overrides
+                .iter()
+                .map(|(ip, socket)| (*ip, *socket)),
         }
     }
 
--- a/client-core/src/data_store.rs
+++ b/client-core/src/data_store.rs
@@ -1,8 +1,9 @@
 use anyhow::{bail, Error};
+use colored::Colorize;
 use innernet_shared::{
     chmod, ensure_dirs_exist, Cidr, Endpoint, IoErrorContext, Peer, WrappedIoError,
 };
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use std::{
     collections::BTreeMap,
     fs::{File, OpenOptions},
@@ -12,30 +13,13 @@ use std::{
 };
 use wireguard_control::InterfaceName;
 
-/// File-backed storage of [`Peer`]s and [`Cidr`]s.
-///
-/// The file is stored at [`Self::get_path()`].
 #[derive(Debug)]
-pub struct DataStore {
+struct FileBackedData<T> {
     file: File,
-    contents: Contents,
+    contents: T,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
-#[serde(tag = "version")]
-enum Contents {
-    #[serde(rename = "1")]
-    V1 {
-        peers: Vec<Peer>,
-        cidrs: Vec<Cidr>,
-
-        /// This field was added in innernet 1.8, but the change is backwards-compatible, we thus kept it in v1
-        #[serde(default)]
-        peer_endpoint_overrides: BTreeMap<IpAddr, Endpoint>,
-    },
-}
-
-impl DataStore {
+impl<T: Default + DeserializeOwned> FileBackedData<T> {
     fn open_with_path<P: AsRef<Path>>(path: P, create: bool) -> Result<Self, WrappedIoError> {
         let path = path.as_ref();
         let is_existing_file = path.exists();
@@ -55,18 +39,75 @@ impl DataStore {
 
         let mut json = String::new();
         file.read_to_string(&mut json).with_path(path)?;
-        let contents = serde_json::from_str(&json).unwrap_or_else(|_| Contents::V1 {
-            peers: vec![],
-            cidrs: vec![],
-            peer_endpoint_overrides: BTreeMap::new(),
-        });
+        let contents = serde_json::from_str(&json).unwrap_or_else(|_| T::default());
 
         Ok(Self { file, contents })
     }
+}
 
+#[derive(Debug)]
+pub struct DataStore {
+    /// File-backed storage of [`Peer`]s and [`Cidr`]s.
+    ///
+    /// The file is stored at [`Self::get_peers_and_cidrs_path()`].
+    peers_and_cidrs: FileBackedData<PeersAndCidrs>,
+
+    /// File-backed storage of [`PeerEndpointOverrides`]s.
+    ///
+    /// The file is stored at [`Self::get_peer_endpoint_overrides_path()`].
+    peer_endpoint_overrides: FileBackedData<PeerEndpointOverrides>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(tag = "version")]
+enum PeersAndCidrs {
+    #[serde(rename = "1")]
+    V1 { peers: Vec<Peer>, cidrs: Vec<Cidr> },
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(tag = "version")]
+enum PeerEndpointOverrides {
+    #[serde(rename = "1")]
+    V1 {
+        /// This field was added in innernet 1.8, but the change is
+        /// backwards-compatible, we thus kept it in v1
+        #[serde(default)]
+        peer_endpoint_overrides: BTreeMap<IpAddr, Endpoint>,
+    },
+}
+
+impl Default for PeersAndCidrs {
+    fn default() -> Self {
+        Self::V1 {
+            peers: vec![],
+            cidrs: vec![],
+        }
+    }
+}
+
+impl Default for PeerEndpointOverrides {
+    fn default() -> Self {
+        Self::V1 {
+            peer_endpoint_overrides: BTreeMap::new(),
+        }
+    }
+}
+
+impl DataStore {
     /// Produce the path "`data_dir`/`interface`.json".
-    pub fn get_path(data_dir: &Path, interface: &InterfaceName) -> PathBuf {
+    pub fn get_peers_and_cidrs_path(data_dir: &Path, interface: &InterfaceName) -> PathBuf {
         data_dir.join(interface.to_string()).with_extension("json")
+    }
+
+    /// Produce the path "`data_dir`/`interface`.peer-endpoint-overrides.json".
+    pub fn get_peer_endpoint_overrides_path(data_dir: &Path, interface: &InterfaceName) -> PathBuf {
+        let file_name = format!("{interface}.peer-endpoint-overrides");
+        data_dir.join(file_name).with_extension("json")
+    }
+
+    pub fn peers_and_cidrs_file_exists(data_dir: &Path, interface: &InterfaceName) -> bool {
+        Self::get_peers_and_cidrs_path(data_dir, interface).exists()
     }
 
     fn _open(
@@ -75,7 +116,19 @@ impl DataStore {
         create: bool,
     ) -> Result<Self, WrappedIoError> {
         ensure_dirs_exist(&[data_dir])?;
-        Self::open_with_path(Self::get_path(data_dir, interface), create)
+        let peers_and_cidrs = FileBackedData::open_with_path(
+            Self::get_peers_and_cidrs_path(data_dir, interface),
+            create,
+        )?;
+        let peer_endpoint_overrides = FileBackedData::open_with_path(
+            Self::get_peer_endpoint_overrides_path(data_dir, interface),
+            create,
+        )?;
+
+        Ok(Self {
+            peers_and_cidrs,
+            peer_endpoint_overrides,
+        })
     }
 
     /// Open the data store residing at "`data_dir`/`interface`.json".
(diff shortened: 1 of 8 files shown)


client-core/src/data_store.rs
```diff
@@ -140,13 +150,53 @@ impl DataStore {
...
+    }
+
     fn set_cidrs(&mut self, new_cidrs: Vec<Cidr>) {
         match &mut self.contents {
             Contents::V1 { ref mut cidrs, .. } => *cidrs = new_cidrs,
         }
     }
 
-    fn write(&mut self) -> Result<(), io::Error> {
+    pub fn write(&mut self) -> Result<(), io::Error> {
```
bschwind: This was probably private for a reason, though I'm not exactly sure what that reason is.

client/src/main.rs
```diff
@@ -781,6 +789,52 @@ fn override_endpoint(
...
+                .apply(interface, opts.network.backend)?;
+        } else {
+            log::info!(
+                "unsetting endpoint override for peer IP {} with endpoint",
+                peer.ip
+            );
+            data_store.unset_endpoint_for_peer(peer.ip);
+        }
+
+        // TODO(bschwind) - Should we do NAT traversal too?
```
bschwind: I _think_ we don't need NAT traversal, since we're explicitly setting just one endpoint for the peer and it applies as soon as we run `DeviceUpdate::new(...).apply()` above.

shared/src/prompts.rs
```diff
@@ -567,3 +568,78 @@ pub fn input_external_endpoint(
...
+/// Bring up a prompt to override the endpoint for an existing peer.
+/// Returns the peer and desired endpoint.
+pub fn override_peer_endpoint_prompt(
+    peers: &[Peer],
+    peer_endpoint_overrides: &HashMap<IpAddr, Endpoint>,
+    args: &OverridePeerEndpointOpts,
+) -> Result<Option<(Peer, Option<Endpoint>)>, Error> {
+    let eligible_peers = peers
+        .iter()
+        .filter(|p| &*p.name != "innernet-server")
```
bschwind, line 581: _Do_ we want to filter out the `innernet-server`? I suppose it could be helpful in some scenarios.

client-core/src/data_store.rs
```diff
@@ -21,7 +25,12 @@ pub struct DataStore {
...
 enum Contents {
     #[serde(rename = "1")]
-    V1 { peers: Vec<Peer>, cidrs: Vec<Cidr> },
+    V1 {
+        peers: Vec<Peer>,
+        cidrs: Vec<Cidr>,
+        #[serde(default)]
+        local_endpoint_overrides: HashMap<IpAddr, Endpoint>,
+    },
 }
```
strohel, line 25: Oh I wasn't aware we have this little neat versioning. But ack that there's no need to bump the version as this id backwards compatible.

Maybe just a comment explaining that it changed over time?
```rust
/// This field was added in innernet 1.8, but the change is backwards-compatible, we thus kept it in v1
```

client-core/src/data_store.rs
```diff
@@ -21,7 +25,12 @@ pub struct DataStore {
 #[serde(tag = "version")]
 enum Contents {
     #[serde(rename = "1")]
-    V1 { peers: Vec<Peer>, cidrs: Vec<Cidr> },
+    V1 {
+        peers: Vec<Peer>,
+        cidrs: Vec<Cidr>,
+        #[serde(default)]
+        local_endpoint_overrides: HashMap<IpAddr, Endpoint>,
```
strohel: I understand the intent with "local", but it feels rather confusing with `--override-endpoit`, which is, in some sense, an endpoint of ourselves, the local peer.

I'd call it `peer_endpoint_overrides` -- everywhere.

strohel: We should use BTreeMap: HashMap would shuffle the entries on every invocation.

client-core/src/data_store.rs
```diff
@@ -140,13 +150,53 @@ impl DataStore {
         }
     }
 
+    pub fn local_endpoint_overrides(&self) -> &HashMap<IpAddr, Endpoint> {
+        match &self.contents {
+            Contents::V1 {
+                local_endpoint_overrides,
+                ..
+            } => local_endpoint_overrides,
+        }
```
strohel, line 146: Implementation nit: little helpers `fn peer_endpont_overrides[_mut](&[mut] self) -> &[mut] BTreeMap<..>` would save us from typing the match in every method.

client-core/src/data_store.rs
strohel: > This was probably private for a reason, though I'm not exactly sure what that reason is.

Looks like there was just never a need to, feels safe to do.

client/src/main.rs
```diff
@@ -781,6 +790,52 @@ fn override_endpoint(
     Ok(())
 }
 
+fn override_peer_endpoint(
+    interface: &InterfaceName,
+    opts: &Opts,
+    sub_opts: OverridePeerEndpointOpts,
+) -> Result<(), Error> {
+    let _config = InterfaceConfig::from_interface(&opts.config_dir, interface)?;
```
strohel: Hmm, why this line exists? Just to check that the innernet config file exists (otherwise I guess this would happily work on just the data store?)

Sounds reasonable, but I'd stick a comment. I guess this might have been copied form another methods.

client/src/main.rs
strohel: Makes sense.

client/src/main.rs
```diff
@@ -781,6 +790,52 @@ fn override_endpoint(
...
+
+            DeviceUpdate::new()
+                .add_peer(PeerConfigBuilder::new(&peer_pub_key).set_endpoint(socket_addr))
+                .apply(interface, opts.network.backend)?;
+        } else {
+            log::info!(
+                "unsetting endpoint override for peer IP {} with endpoint",
+                peer.ip
+            );
+            data_store.unset_endpoint_for_peer(peer.ip);
```
strohel: I see we just remove the override from the data store, and don't touch the actual interface. I'd mention in the log, that the "normal" endpoint/NAT traversal will be restored on the next `innernet fetch` run (perhaps performed by the daemon).

client/src/main.rs
```diff
@@ -781,6 +790,52 @@ fn override_endpoint(
     Ok(())
 }
 
+fn override_peer_endpoint(
+    interface: &InterfaceName,
+    opts: &Opts,
+    sub_opts: OverridePeerEndpointOpts,
+) -> Result<(), Error> {
```
strohel, line 799: There seems to be a race condition with this approach. Suppose you have an (1) `innernet up %i --daemon --interval 60` running (the shipped systemd service), and then you call (2) `innernet --override-peer-endpoint`.

(2) writes to the store and actually sets the endpoint on the interface
(1) Might be inside the `fetch` function for the given interface (this is quite likely, with nat traversal it takes tens of seconds). It only reads the data store at the beginning and overwrites it at the end. So I think it will happily drop the peer endpoint override from both the data store and the interface. 💥

I guess in (1) we can do a trick like "only write the data store file at the end if it hasn't changed in the mean time". But feels a bit tricky.

client/src/main.rs
```diff
@@ -963,19 +1018,26 @@ fn print_interface(device_info: &Device, short: bool) -> Result<(), Error> {
...
 }
 
-fn print_peer(peer: &PeerState, short: bool, level: usize) {
+fn print_peer(peer: &PeerState, short: bool, level: usize, data_store: &DataStore) {
     let pad = level * 2;
     let PeerState { peer, info } = peer;
+
+    let endpoint_has_local_override = data_store.endpoint_override_for_peer(peer.ip).is_some();
+    let endpoint_override_msg = if endpoint_has_local_override {
+        " (local endpoint override)"
```
strohel: Good call mentioning this in the output! That was one of my thoughts too.

I'd probably say
```suggestion
        " (endpoint overridden locally)"
```

Otherwise it may sound like that is "local endpoint" rather than "local override".

client/src/main.rs
```diff
@@ -963,19 +1018,26 @@ fn print_interface(device_info: &Device, short: bool) -> Result<(), Error> {
...
         let connected = info
             .map(|info| info.is_recently_connected())
             .unwrap_or_default();
-
         let is_you = info.is_none();
 
         println_pad!(
             pad,
-            "| {} {}: {} ({}{}…)",
+            "| {} {}: {} ({}{}…){}",
```
strohel: I see that would be a mixed usage, but I still find that more readable? Similar below, where 2 variables can be inlined.
```suggestion
            "| {} {}: {} ({}{}…){endpoint_override_msg}",
```

shared/src/prompts.rs
```diff
@@ -567,3 +568,78 @@ pub fn input_external_endpoint(
...
+        // If we're unsetting, filter eligible_peers to just be peers that have
+        // an override already set.
+        .filter(|p| !args.unset || peer_endpoint_overrides.contains_key(&p.ip))
+        .collect::<Vec<_>>();
+
+    if args.unset && eligible_peers.is_empty() {
+        return Err(anyhow!("No peers have an override endpoint set"));
+    }
+
+    let peer = if let Some(ref name) = args.name {
```
strohel: Supernit I think no need to use ref:

```suggestion
    let peer = if let Some(name) = &args.name {
```

Otherwise this, but I _think_ we don't need to
```suggestion
    let peer = if let Some(name) = args.name.as_ref() {
```

shared/src/prompts.rs
strohel: If you want to override the endpoint of the server, you can edit it in `/etc/innernet/<iface>.conf`.

But, is this the correct way to filter-out the server? Maybe it can have a different name? I thought I've seen server being filtered in other places in code, might have used a different method.

shared/src/prompts.rs
```diff
@@ -567,3 +568,78 @@ pub fn input_external_endpoint(
...
+        .iter()
+        .filter(|p| &*p.name != "innernet-server")
+        // If we're unsetting, filter eligible_peers to just be peers that have
+        // an override already set.
+        .filter(|p| !args.unset || peer_endpoint_overrides.contains_key(&p.ip))
+        .collect::<Vec<_>>();
+
+    if args.unset && eligible_peers.is_empty() {
+        return Err(anyhow!("No peers have an override endpoint set"));
+    }
```
strohel: I'd move this check to
```rust
    } else {
        let message = if args.unset {
```

When `args.name` is specified _and_ `args.unset == true` _and_ that peer doesn't have peer EP override, I'd return an `Ok` value, just mentioning it wasn't set in the first place. That makes the command much more usable in scripts.

client-core/src/data_store.rs
```diff
@@ -140,13 +150,53 @@ impl DataStore {
...
+    pub fn endpoint_override_for_peer(&self, peer_ip: IpAddr) -> Option<Endpoint> {
+        match &self.contents {
+            Contents::V1 {
+                local_endpoint_overrides,
+                ..
+            } => local_endpoint_overrides.get(&peer_ip).cloned(),
+        }
+    }
+
+    pub fn override_endpoint_for_peer(&mut self, peer_ip: IpAddr, endpoint: Endpoint) {
```
mbernat: I got a bit confused by 'endpoint override' vs 'override endpoint', so I vote for the more boring name.

```suggestion
    pub fn set_endpoint_override_for_peer(&mut self, peer_ip: IpAddr, endpoint: Endpoint) {
```

client-core/src/data_store.rs
```diff
@@ -140,13 +150,53 @@ impl DataStore {
...
+            Contents::V1 {
+                local_endpoint_overrides,
+                ..
+            } => {
+                local_endpoint_overrides.insert(peer_ip, endpoint);
+            },
+        }
+    }
+
+    pub fn unset_endpoint_for_peer(&mut self, peer_ip: IpAddr) {
```
mbernat:
```suggestion
    pub fn unset_endpoint_override_for_peer(&mut self, peer_ip: IpAddr) {
```

strohel (dismissed):
Oh neat, this will be very handy!

Some technicalities, one bigger one - a race condition between fetch (perhaps called from the daemon mode) and this command.

FWIW I originally thought of a more naive approach where the overrides wouldn't be persistent: we'd have to specify them for all `up` / `fetch` invocations. That would be of course much less ergonomic, so this is better (but we have to deal with the race condition).

client/src/main.rs
bschwind: Oops, this was just a copy-paste and I subsequently forgot to remove it.

client/src/main.rs
bschwind: Good catch, that's a bit unfortunate. We could perhaps store it in a separate file.

Otherwise I would maybe say something like an sqlite database would be the way to go for storing innernet client state in the future. I'll think more about this.

strohel: Yeah using a separate file was something that also appeared to me while having lunch. I think that should work!

SQLite may indeed be the future, but much bigger scope, yeah.

client/src/main.rs
```diff
@@ -781,6 +790,52 @@ fn override_endpoint(
...
+                endpoint
+            );
+            data_store.override_endpoint_for_peer(peer.ip, endpoint.clone());
+
+            let socket_addr = endpoint.resolve()?;
+            let peer_pub_key = Key::from_base64(&peer.public_key).unwrap();
+
+            DeviceUpdate::new()
+                .add_peer(PeerConfigBuilder::new(&peer_pub_key).set_endpoint(socket_addr))
+                .apply(interface, opts.network.backend)?;
```
strohel: Now I see that this `apply()` will also race with the `fetch()` in the client daemon. We will unconditionally set the endpoint here, but `fetch()` may be doing NAT traversal for this peer, and will continue doing so. At least the next fetch will restore the override from store, but still.

I think it may be most predictable not to touch the interface here, just tell users in the output that the change will be effective with the next `fetch` invocation.

client/src/main.rs
mbernat, line 799: It would be nice to also provide this as a `client-core` API, now that we have that crate.

mbernat (commented):
Just some quick comments.

strohel (commented):
Forgot to react to your questions earlier:
> One open question: if the specified endpoint is not reachable, should we allow NAT traversal to run and potentially set the endpoint to something different which is reachable? Or should it remain fixed on that specified endpoint override no matter what?

I think setting an override endpoint should disable NAT traversal for that peer, yes.

Another question: innernet client has a behavior that it doesn't touch an endpoint of a peer if there has been handshake in last 3 minutes with that endpoint. Shall this behavior apply also when you override endpoint? Perhaps not? Shall we at least log the fact that we may be breaking the connection?

shared/src/prompts.rs
bschwind: I'm confused, are you saying you would move the highlighted code to the `else` block you quoted? When we're in that `else` block, that means `args.name` was not specified.

If you can, I'd prefer a code suggestion, or if github doesn't let you write one nicely, at least a more full suggestion in a code block.

client-core/src/data_store.rs
```diff
@@ -28,8 +28,10 @@ enum Contents {
     V1 {
         peers: Vec<Peer>,
         cidrs: Vec<Cidr>,
+
+        /// This field was added in innernet 1.8, but the change is backwards-compatible, we thus kept it in v1
```
strohel: Sorry let's keep this at max 100 cols.

client-core/src/data_store.rs
```diff
@@ -150,43 +152,38 @@ impl DataStore {
...
-    pub fn local_endpoint_overrides(&self) -> &HashMap<IpAddr, Endpoint> {
-        match &self.contents {
-            Contents::V1 {
-                local_endpoint_overrides,
-                ..
-            } => local_endpoint_overrides,
-        }
+    pub fn peer_endpoint_overrides(&self) -> &BTreeMap<IpAddr, Endpoint> {
+        self.inner_peer_endpoint_overrides()
     }
```
strohel: Nit no need for the inner_ method, can be made public right away.

shared/src/prompts.rs
```diff
@@ -567,3 +568,78 @@ pub fn input_external_endpoint(
...
+
+        let (peer_index, _) = select(
+            message,
+            &eligible_peers
+                .iter()
+                .map(|ep| ep.name.clone())
+                .collect::<Vec<_>>(),
+        )?;
+        eligible_peers[peer_index].clone()
+    };
```
strohel, line 616: Formulating my earlier comment as a suggestion:
```suggestion
pub fn override_peer_endpoint_prompt(
    peers: &[Peer],
    peer_endpoint_overrides: &BTreeMap<IpAddr, Endpoint>,
    args: &OverridePeerEndpointOpts,
) -> Result<Option<(Peer, Option<Endpoint>)>, Error> {
    let eligible_peers = peers
        .iter()
        .filter(|p| &*p.name != "innernet-server")
        // If we're unsetting, filter eligible_peers to just be peers that have
        // an override already set.
        .filter(|p| !args.unset || peer_endpoint_overrides.contains_key(&p.ip))
        .collect::<Vec<_>>();

    let peer = if let Some(name) = &args.name {
        let Some(peer) = eligible_peers.into_iter().find(|p| &p.name == name) else {
            return if args.unset && peers.iter().find(|p| &p.name == name).is_some() {
                log::info("Peer '{name}' does not have an override set.");
                Ok(None)
            } else {
                Err(anyhow!("Peer '{name}' does not exist."))
            };
        };
        peer
    } else {
        let message = if args.unset {
            if eligible_peers.is_empty() {
                return Err(anyhow!("No peers have an override endpoint set"));
            }

            "Peer endpoint override to unset"
        } else {
            "Peer endpoint to override"
        };

        let (peer_index, _) = select(
            message,
            &eligible_peers
                .iter()
                .map(|ep| ep.name.clone())
                .collect::<Vec<_>>(),
        )?;
        eligible_peers[peer_index].clone()
    };
```

client/src/main.rs
bschwind: If you're cool with a separate file, I think that's really the only approach we can do with how things are currently set up. The DataStore `write()` completely wipes the file, so you can't really avoid race conditions if you also allow concurrent processes to access the file.

strohel: > If you're cool with a separate file

Definitely! Something like `/var/lib/innernet/<interface>.peer-endpoint-overrides.json`. (or the same with underscores if people prefer that)

`<interface>` _can_ contain dots, but if can be max 15 chars long, hence it cannot clash. It _cannot_ contain spaces, but using space as a separator doesn't feel good.
https://github.com/tonarino/innernet/blob/fe544aa52e27d968cd71484a62e92802ab33385c/wireguard-control/src/device.rs#L139-L168

shared/src/prompts.rs
bschwind: This is how it's done in other places, so I just kept the convention going.

client/src/main.rs
bschwind: Okay, I pushed a version which uses a separate file. I don't _love_ it but I think it'll work.

bschwind: Oh here's a fun bug:

```rust
/// Produce the path "`data_dir`/`interface`.peer-endpoint-overrides.json".
pub fn get_peer_endpoint_overrides_path(data_dir: &Path, interface: &InterfaceName) -> PathBuf {
    let file_name = format!("{interface}.peer-endpoint-overrides");
    data_dir.join(file_name).with_extension("json")
}
```

Calling `.with_extension()` here _replaces` the `.peer-endpoint-overrides` string we added above, resulting in `<INTERFACE_NAME>.json`, which means you'll end up overwriting the main peers and CIDRs file.

bschwind: We can use [`with_added_extension`](https://doc.rust-lang.org/std/path/struct.Path.html#method.with_added_extension) though it's relatively new, introduced in Rust 1.91.

mbernat: Or just this, since we're not using `PathBuf` for the first extension either.

```
let file_name = format!("{interface}.peer-endpoint-overrides.json");
```

client-core/src/data_store.rs
```diff
@@ -55,27 +39,96 @@ impl DataStore {
...
+    /// File-backed storage of [`Peer`]s and [`Cidr`]s.
+    ///
+    /// The file is stored at [`Self::get_peers_and_cidrs_path()`].
+    peers_and_cidrs: FileBackedData<PeersAndCidrs>,
+
+    /// File-backed storage of [`PeerEndpointOverrides`]s.
+    ///
+    /// The file is stored at [`Self::get_peer_endpoint_overrides_path()`].
+    peer_endpoint_overrides: FileBackedData<PeerEndpointOverrides>,
+}
```
strohel: Ah, interesting approach, didn't occur to me we could group these 2.

I'll give a a thought. To me, `PeersAndCidrs` is mostly "a cache of what the server sends to us", while `PeerEndpointOverrides` is more a config.

But there already is config, `/etc/innernet/<interface>.conf`.

I first thought that we shouldn't be touching it, but now I realized that `innernet set-listen-port` does touch it.

On that wave, perhaps `innernet override-peer-endpoint` should really persist its values in `/etc/innernet/<interface>.conf`? Perhaps in a new section?

I haven't built any strong opinions yet, just sharing my line of thought quickly.

bschwind: That might make more sense. A local sqlite DB still makes the most sense as the "proper" way to do it.

It feels a little sketchy to be doing a lot of operations in the main file that holds your private key, but it's true that we already modify the listen port in this file.

I think we have most of the logic in place in this PR to modify the peer endpoint overrides, so we just need to pick a place to store it and then we should be good to go.

client/src/main.rs
```diff
@@ -963,19 +1013,26 @@ fn print_interface(device_info: &Device, short: bool) -> Result<(), Error> {
     Ok(())
 }
 
-fn print_peer(peer: &PeerState, short: bool, level: usize) {
+fn print_peer(peer: &PeerState, short: bool, level: usize, data_store: &DataStore) {
```
strohel: Sorry another late comment, nit: perhaps this should take `has_endpoint_override: bool` or `endpoint_override: Option<...>` rather than the whole `data_store`/`config`, so that it is more "local"?

bschwind:
I've moved the storage of endpoint overrides into `InterfaceConfig`, now I just need to follow up on a few more review comments (including putting this logic in `client-core`) and then hopefully it will be good for another review.

I had to do some nasty type gymnastics in the `show()` function in `main.rs`, maybe someone has a better way of doing it.

client/src/main.rs
bschwind: This data ended up moving to `InterfaceConfig`, so no extra files are introduced.

client/src/main.rs
bschwind: I think it's better to apply it immediately. If you don't happen to have fetch running continuously in the background (I don't, for example) then I would expect the change to apply when I run it.

Sure it might get changed once from NAT traversal, but like you said the next fetch will correct it.

client/src/main.rs
bschwind: Done!

bschwind:
> Another question: innernet client has a behavior that it doesn't touch an endpoint of a peer if there has been handshake in last 3 minutes with that endpoint. Shall this behavior apply also when you override endpoint? Perhaps not? Shall we at least log the fact that we may be breaking the connection?

I think we should just apply the override. No need to log in my opinion, as someone running this command probably understands what they're doing.

shared/src/prompts.rs
bschwind: @strohel can you check out the branch locally and come up with a version you like that compiles? This one doesn't, and the code gets kind of ugly to make it compile.

client-core/src/interface.rs
```diff
@@ -233,6 +234,19 @@ pub fn fetch(
...
+            endpoint_override
+        );
+
+        if let Some(peer) = peers.iter_mut().find(|p| p.ip == *peer_ip) {
+            peer.endpoint = Some(endpoint_override.clone());
+        }
+    }
+
     let device = Device::get(interface, network_opts.backend)?;
     let modifications = device.diff(&peers);
```
strohel, line 248: This transitively calls PeerDiff::new() -> PeerDiff::peer_config_builder() -> https://github.com/tonarino/innernet/blob/cd105cc7c1ee7a306febd3c47c7fa3b455ed8f0d/shared/src/types.rs#L733-L749

So I think this intention isn't fulfilled:
> > Another question: innernet client has a behavior that it doesn't touch an endpoint of a peer if there has been handshake in last 3 minutes with that endpoint. Shall this behavior apply also when you override endpoint? Perhaps not? Shall we at least log the fact that we may be breaking the connection?
> 
> I think we should just apply the override. No need to log in my opinion, as someone running this command probably understands what they're doing.

OTOH:
- `innernet override-peer-endpoit` does replace the peer endpoint right away
- the very first `innernet up` should see `info.is_recently_connected() == false`
- ...so this should probably work out?

When WireGuard sees a valid incoming packet from `<source endpoint>`, it happily replaces its endpoint on file with `<source endpoint>`. So maybe trying this for an already established connection is futile.

Anyways, nothing to be done here. I've added an item to testing to check the behavior.

client/src/main.rs
```diff
@@ -781,6 +792,56 @@ fn override_endpoint(
...
+            );
+
+            innernet_client_core::unset_endpoint_override_for_peer(
+                &opts.config_dir,
+                interface,
+                &mut config,
+                &peer,
+            )?;
+
+            log::info!("normal peer endpoint/NAT traversal will be restored on the next call to 'innernet fetch'");
```
strohel, line 833: Nit: this is 116 cols, let's break it. It seem we donẗ have `reformat_strings` enabled for innernet, I think eventually we should.

client/src/main.rs
```diff
@@ -847,18 +908,24 @@ fn show(opts: &Opts, short: bool, tree: bool, interface: Option<Interface>) -> R
...
+            let fallible_things = || {
+                let device =
+                    Device::get(&name, opts.network.backend).with_str(name.as_str_lossy())?;
+                let config = InterfaceConfig::from_interface(&opts.config_dir, &name)?;
+
+                Ok((device, data_store, config))
+            };
+
+            Some(fallible_things())
         })
```
strohel, line 922: This compiles:
```suggestion
        .map(|name| -> Result<Option<_>, Error> {
            match DataStore::open(&opts.data_dir, &name) {
                Ok(store) => {
                    let device =
                        Device::get(&name, opts.network.backend).with_str(name.as_str_lossy())?;
                    let config = InterfaceConfig::from_interface(&opts.config_dir, &name)?;

                    Ok(Some((device, store, config)))
                },
                // Skip WireGuard interfaces that aren't managed by innernet.
                Err(e) if e.kind() == io::ErrorKind::NotFound => Ok(None),
                // Error on interfaces that *are* managed by innernet but are not readable.
                Err(e) => Err(e.into()),
            }
        })
        // Filter out Ok(None).
        .filter_map(Result::transpose)
```

shared/src/interface_config.rs
```diff
@@ -133,8 +148,16 @@ impl InterfaceConfig {
...
+        &self.peer_endpoint_overrides
+    }
+
+    pub fn set_endpoint_override_for_peer(&mut self, peer_ip: IpAddr, endpoint: Endpoint) {
+        self.peer_endpoint_overrides.insert(peer_ip, endpoint);
+    }
+
+    pub fn unset_endpoint_override_for_peer(&mut self, peer_ip: IpAddr) {
+        self.peer_endpoint_overrides.remove(&peer_ip);
     }
```
strohel, line 161: Optional: a single method that takes `Option<Endpoint>` might be a mit more ergonomic for callers here and in innrnet_client_core, client, but up to you.

shared/src/prompts.rs
strohel: The suggestion was missing one `!` and one `.clone()`, this one compiles:
```suggestion
pub fn override_peer_endpoint_prompt(
    peers: &[Peer],
    peer_endpoint_overrides: &BTreeMap<IpAddr, Endpoint>,
    args: &OverridePeerEndpointOpts,
) -> Result<Option<(Peer, Option<Endpoint>)>, Error> {
    let eligible_peers = peers
        .iter()
        .filter(|p| &*p.name != "innernet-server")
        // If we're unsetting, filter eligible_peers to just be peers that have
        // an override already set.
        .filter(|p| !args.unset || peer_endpoint_overrides.contains_key(&p.ip))
        .collect::<Vec<_>>();

    let peer = if let Some(name) = &args.name {
        let Some(peer) = eligible_peers.into_iter().find(|p| &p.name == name) else {
            return if args.unset && peers.iter().find(|p| &p.name == name).is_some() {
                log::info!("Peer '{name}' does not have an override set.");
                Ok(None)
            } else {
                Err(anyhow!("Peer '{name}' does not exist."))
            };
        };
        peer.clone()
    } else {
        let message = if args.unset {
            if eligible_peers.is_empty() {
                bail!("No peers have an override endpoint set");
            }

            "Peer endpoint override to unset"
        } else {
            "Peer endpoint to override"
        };

        let (peer_index, _) = select(
            message,
            &eligible_peers
                .iter()
                .map(|ep| ep.name.clone())
                .collect::<Vec<_>>(),
        )?;
        eligible_peers[peer_index].clone()
    };
```

strohel (commented):
Nice! Just some small technicalities left, but I think this is now good/compete on the fundamental level! I've also added some testing checkboxes to the description.

client/src/main.rs
bschwind: Let's leave it as is, and change the rustfmt rules if that's what we want.

client/src/main.rs
bschwind: Nice one, thanks!

shared/src/interface_config.rs
bschwind: I think it's about the same amount of ergonomics, as the calling code still needs some conditional statements to decide what to log

strohel (approved):
Nice! Code LGTM, now onto testing. Tentative checklist in the description.

bschwind:
@goodhoko and I tested this a bit, the first test step already revealed an issue.

When you do `DeviceUpdate::apply()` when the interface is down, it will actually bring it back up into a partial state, but without doing the full setup such as adding the innernet server as a peer. On subsequent calls to `innernet up`, our "is the interface up" logic returns `true`, so it doesn't go through the full `up` routine, and immediately tries to fetch from the server. That fetch times out, of course.

So we added a change to only apply the peer endpoint override if the interface is currently up.

bschwind:
We tested the last point, but kind of in reverse. We had our test peers using the relay to communicate. We then both unset the local peer endpoint overrides, but wireguard very aggressively continued using the relay endpoints because they were still connected.

Aside from an `innernet down` and `innernet up`, or a machine reboot, I'm not sure how we would ever convince the peers to stop using the relay if we really wanted to.

strohel (approved):
Latest changes LGTM.

> We had our test peers using the relay to communicate. We then both unset the local peer endpoint overrides, but wireguard very aggressively continued using the relay endpoints because they were still connected.
> 
> Aside from an `innernet down` and `innernet up`, or a machine reboot, I'm not sure how we would ever convince the peers to stop using the relay if we really wanted to.

Yeah, the "working" endpoints seem to be really sticky.

If desperate we can introduce firewall rules (on client or relay level), but yeah, that's meh.

mbernat:
> Aside from an `innernet down` and `innernet up`...

Would it make sense to have a dedicated restart command for doing down and up in succession so that we can avoid using `tonari-rescue`? The same command could also be invoked when unsetting the endpoint overrides.

bschwind:
I just fixed the merge conflicts with `print_peer`, code might be worth another look.

strohel:
> > Aside from an `innernet down` and `innernet up`...
> 
> Would it make sense to have a dedicated restart command for doing down and up in succession so that we can avoid using `tonari-rescue`? The same command could also be invoked when unsetting the endpoint overrides.

I don't think (the equivalent of) doing `innernet down` and `innernet up` in quick succession would help. You want the other side to "forget" your working endpoint, I think in general you'd need to wait 4 minutes between down and up: 3 minutes so that peer's innnernet doesn't consider the endpoint online, and one extra minute to let `innernet fetch` kick in and set some "official" endpoint.

strohel (approved):
Still LGTM.

goodhoko:
> You want the other side to "forget" your working endpoint, I think in general you'd need to wait 4 minutes between down and up:

Yup. You can also bring the network down for a shorter while if you do it for both peers at the same time. That way, wireguard forgets the relay endpoint *and* there are no packets coming that would "remind" it.

skywhale:
@bschwind let's merge this 🙏 could you also cut a new release?

bschwind:
Okay, I'll go ahead and merge this and cut a release today.

mbernat:
@bschwind Just FYI: `main` has some breaking changes in `innernet-shared`, so it would be best to release a 2.0 in case that crate has any users.

See this thread where we briefly discussed the alternatives with @strohel: https://github.com/tonarino/innernet/pull/382#discussion_r2857927251

PR #397, opened 2026-06-02
Include innernet_client_core in BASE_MODULES for logging

bschwind:
One thing I noticed when working on #395

When you run something simple like `innernet fetch`, it prints basically nothing. This is because a lot of the logic was moved to `innernet_client_core`, and that "module" gets filtered out by our logging logic.

This fixes it, though I also question whether `innernet_client_core` should even be in charge of logging user-facing messages like this. Usually the application layer that is using the "core" library would be the one to print those messages.

### DIFF
--- a/client/src/util.rs
+++ b/client/src/util.rs
@@ -7,7 +7,7 @@ use std::{ffi::OsStr, io, path::Path, time::Duration};
 static LOGGER: Logger = Logger;
 struct Logger;
 
-const BASE_MODULES: &[&str] = &["innernet", "innernet_shared"];
+const BASE_MODULES: &[&str] = &["innernet", "innernet_client_core", "innernet_shared"];
 
 fn target_is_base(target: &str) -> bool {
     BASE_MODULES


strohel (approved):
Oh thanks! Me and @mbernat were wondering yesterday why `innernet` no longer prints basically anything, this explains it.

As of the more general question I'll leave it as @mbernat's food for thought.

Agreed that it should be usually the end applications that log. OTOH doing that in client_core is currently pragmatic, and "library logs even if I don't want it to" problem has a solution in form of log targets.

mbernat (approved):
LGTM, thanks!!!

mbernat:
> I also question whether innernet_client_core should even be in charge of logging user-facing messages like this. Usually the application layer that is using the "core" library would be the one to print those messages.

I agree, a library shouldn't be doing this. But innernet likes to be very chatty in some places, logging many things as it goes along, so you'd have to pass those messages from the core up the call chain. At that point might be easier to just let it log everything and let the application filter things out.

A separate question is why innernet is so chatty in the first place, especially in places where it's not really needed, while staying completely silent in many other places, like many error messages. I'd like to resolve this question first before picking a particular logging strategy.

PR #398, opened 2026-06-16
Add -v/--verbose flag to print public keys in full

### LINKED ISSUE #344: Option for `innernet show` not to ellipsize the wirguard key
Currently, when one does `innernet show`, the output looks like
```
  peer: joe (fk8Hnjnfdk...)
    ip: 1.2.3.4
    endpoint: 192.168.1.2:51951
    last handshake: 2 hours, 30 minutes ago
    transfer: 0 B received, 1096.56 KiB sent
```

Here `fk8Hnjnfdk` is the start of peer's public key - but not a full one. Sometime the full one is needed, i.e. when one wants to manually override the endpoint using `wg set network peer <full-key> endpoint <x.y.z.w:p>`

It would be nice if `innernet show` accepted some option not to shorten the public key.

CC @bschwind / @jacksongoode if you want to play with innernet a bit.

goodhoko:
Resolves https://github.com/tonarino/innernet/issues/344

Make it work regardless of whether `--tree` or `--short` are used.

Note that for consistency I unified the truncated length with and without `--short` to 10 characters (it was 6 and 10 respectively).

I tested locally that this new option works with all combinations of the existing `show` options.

This was coded in about 20 minutes in an interactive opencode session. Here are all the prompts I used, verbatim:

1. pull latest main, create a new branch called show-verbose
2. add a -v/--verbose option to the innernet show command which makes it print the wireguard public keys in full rather then truncating them as it's doing now. It should work regardless of whether the -t or -s flags are set or not
3. make the the printing logic more dry by pulling the public key (truncated or not) to a variable
4. not that dry ... it's ok to define the variable twice, once for each case of --short
5. or you know what, make as dry as before and simply unify the truncated length to 10 chars regardless of --short
6. verify with cargo test and cargo clippy (all features, all targets...)
7. commit the changes, mention this resolves https://github.com/tonarino/innernet/issues/344, include a summary of the prompts I used in this session, including the main first one verbatim.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -90,6 +90,10 @@ enum Command {
         #[clap(short, long)]
         tree: bool,
 
+        /// Show full public keys (don't truncate)
+        #[clap(short, long)]
+        verbose: bool,
+
         interface: Option<Interface>,
     },
 
@@ -534,7 +538,7 @@ fn list_cidrs(interface: &InterfaceName, opts: &Opts, tree: bool) -> Result<(), 
     if tree {
         let cidr_tree = CidrTree::new(data_store.cidrs());
         colored::control::set_override(false);
-        print_tree(&cidr_tree, &[], 0);
+        print_tree(&cidr_tree, &[], false, 0);
         colored::control::unset_override();
     } else {
         for cidr in data_store.cidrs() {
@@ -839,7 +843,13 @@ fn get_server_capabilities(config: &InterfaceConfig) -> Result<ServerCapabilitie
     }
 }
 
-fn show(opts: &Opts, short: bool, tree: bool, interface: Option<Interface>) -> Result<(), Error> {
+fn show(
+    opts: &Opts,
+    short: bool,
+    tree: bool,
+    verbose: bool,
+    interface: Option<Interface>,
+) -> Result<(), Error> {
     let interfaces = interface.map_or_else(
         || Device::list(opts.network.backend),
         |interface| Ok(vec![*interface]),
@@ -910,17 +920,17 @@ fn show(opts: &Opts, short: bool, tree: bool, interface: Option<Interface>) -> R
 
         if tree {
             let cidr_tree = CidrTree::new(cidrs);
-            print_tree(&cidr_tree, &peer_states, 1);
+            print_tree(&cidr_tree, &peer_states, verbose, 1);
         } else {
             for peer_state in peer_states {
-                print_peer(&peer_state, short, 1);
+                print_peer(&peer_state, short, verbose, 1);
             }
         }
     }
     Ok(())
 }
 
-fn print_tree(cidr: &CidrTree, peers: &[PeerState], level: usize) {
+fn print_tree(cidr: &CidrTree, peers: &[PeerState], verbose: bool, level: usize) {
     println_pad!(
         level * 2,
         "{} {}",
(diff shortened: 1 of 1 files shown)


client/src/main.rs
```diff
@@ -90,6 +90,10 @@ enum Command {
         #[clap(short, long)]
         tree: bool,
 
+        /// Show full public keys (don't truncate)
+        #[clap(short, long)]
+        verbose: bool,
```
goodhoko, line 95: The name of the option is subject to bikeshedding. It feels little silly to have `--verbose` that can be composed with `--short` but any other name I could come up with felt too cumbersome.

client/src/main.rs
```diff
@@ -963,9 +973,16 @@ fn print_interface(device_info: &Device, short: bool) -> Result<(), Error> {
...
-fn print_peer(peer: &PeerState, short: bool, level: usize) {
+fn print_peer(peer: &PeerState, short: bool, verbose: bool, level: usize) {
     let pad = level * 2;
     let PeerState { peer, info } = peer;
+
+    let public_key = if verbose {
+        peer.public_key.clone()
+    } else {
+        format!("{}…", &peer.public_key[..10])
+    };
```
strohel, line 984: Just Rust curiosity question, would this work?

```suggestion
    let public_key = if verbose {
        &peer.public_key
    } else {
        &format!("{}…", &peer.public_key[..10])
    };
```

Maybe only in some new edition? I vaguely recall there were tweaks to temporary variable lifetime extension rules, perhaps right in let expressions. (CC @mbernat in case you're knowledgeable or curious too).

strohel (changes requested):
Oh thanks for picking this up!

Nice try, AI, but `innernet` already has a global `-v` / `--verbose` flag, the intent was that we hook to it. (to its defense, it could have been misled by your initial prompt)

I.e. use https://docs.rs/log/latest/log/macro.log_enabled.html to determine the key length -- no need to pass `verbose` all along.

goodhoko:
This oversight goes completely to me. Damn it, and I really thought it would be this easy - it even worked completely fine. 

Let me close this PR, as there's very little to be reused. I'll open another one.

> I.e. use https://docs.rs/log/latest/log/macro.log_enabled.html to determine the key length -- no need to pass verbose all along.

TBH that sounds like a very ugly hack. Printing is the sole purpose of `print_peer()` and friends and changing this core behavior shouldn't be passed in via an opaque and unrelated side channel (logging has nothing to do with this). An explicit function argument is really not that bad.

client/src/main.rs
goodhoko: It does work, which is very surprising because I vividly remember this wouldn't work in the past.

I even manually wrote the `clone()` here because ai originally put `to_string()` in there.

strohel: > It does work, which is very surprising because I vividly remember this wouldn't work in the past.

Yeah exactly!

strohel: I had to Google that. The mechanism is called temporary lifetime extension: https://doc.rust-lang.org/stable/reference/destructors.html#temporary-lifetime-extension (this documentation is rather cryptic to me tho)

It seems the mechanism was extended in Rust 1.79: https://blog.rust-lang.org/2024/06/13/Rust-1.79.0/#extending-automatic-temporary-lifetime-extension

...so that explains why we had a prior intuition that this doesn't work.

PR #399, opened 2026-06-17
Print full public keys with global --verbose flag; unify truncation length to 10 chars

### LINKED ISSUE #344: Option for `innernet show` not to ellipsize the wirguard key

goodhoko:
See the commits for details. This is a second attempt after #398 

- [Unify public-key truncation length to 10 chars in show](https://github.com/tonarino/innernet/commit/1bccb0bab74e7e3eb968d0c95f3836c1a8403b3f)
    The `--short` mode previously truncated public keys to 6 characters while the default mode used 10. Change both to consistently show 10 characters.
- [Print full public keys with global --verbose flag](https://github.com/tonarino/innernet/commit/ee8e0ebd6fd9167d5df4821c8f03be3d0a4cb843)
    When `innernet -v show` is used (or `-vv` etc.), print peers' public keys in full instead of truncating to 10 characters. Apply this regardless of the `-s`/`--short` or `-t`/`--tree` subcommand flags.

Resolves https://github.com/tonarino/innernet/issues/344

I've tested this locally.

Co-authored-by: opencode (big-pickle) <agent@opencode.ai>
Prompts used:
  - unify the length to which the public key is truncated in the show subcommand when called with --short and without (currently it is 6 and 10 characters respectively), make it 10 unconditionally.
  - commit the changes, write a nice commit message including a disclosure that this was written by an agent and include the prompts used in this session so far
  - make the global innernet -v/--verbose option print the wireguard public keys in full rather then truncating them as it is doing now when calling the show subcommand. It should work regardless of whether the -t or -s show flags are set or not.
  - dont print the key in print_interface
  - make the code more DRY by only defining the key variable once. Also name it public_key
  - verify with cargo clippy and cargo fmt
  - commit the changes. Mention this resolves https://github.com/tonarino/innernet/issues/344. Include a disclosure of ai use including all prompts used since the last commit.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -984,7 +984,7 @@ fn print_peer(peer: &PeerState, short: bool, level: usize) {
             peer.ip.to_string().yellow().bold(),
             peer.name.yellow(),
             if is_you { "you, " } else { "" },
-            &peer.public_key[..6].dimmed(),
+            &peer.public_key[..10].dimmed(),
         );
     } else {
         println_pad!(
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -534,7 +534,7 @@ fn list_cidrs(interface: &InterfaceName, opts: &Opts, tree: bool) -> Result<(), 
     if tree {
         let cidr_tree = CidrTree::new(data_store.cidrs());
         colored::control::set_override(false);
-        print_tree(&cidr_tree, &[], 0);
+        print_tree(&cidr_tree, &[], 0, false);
         colored::control::unset_override();
     } else {
         for cidr in data_store.cidrs() {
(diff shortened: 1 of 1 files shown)


strohel (approved):
Thanks!

I would have used https://docs.rs/log/latest/log/macro.log_enabled.html but you probably have a stronger opinion and I don't care that much.

PR #407, opened 2026-07-16
Fix bug that prevented preserving SELinux context and add tests

jebotz:
(This is an update of #378 which was closed due to inactivity.)

Fixed a small bug in get_temp_path() where temp files were created in the wrong directory. This caused hosts_dir.with_file_name(temp_filename) to produce /hosts.tmpXXX instead of /etc/hosts.tmpXXX, which in turn broke SELinux context preservation.

Also added tests to verify temp file location and SELinux context preservation.

The original PR was approved, but I said to hold off on merging because I still saw a problem.  It turned out that that problem was not in the code but in the rpm packages I had built using the script in the rpm/ subdir.  The code itself is correct and should be merged; in this PR I've only re-synced it with main and added the selinux feature flag to the server cargo file where it was missing.

The rpm package script has several problems (in addition to not enabling the selinux feature when needed), relies on a tool that doesn't seem to be maintained any more (cargo-rpm) and is incompatible with the latest version of rpm (6.x). It is also not invoked in CI, so nobody would notice when it's broken.  It should either be removed or fixed (by using cargo-generate-rpm instead of cargo-rpm).  I can submit a separate PR for that.


strohel (approved):
Thanks for getting back to this, code LGTM!

Ad. clippy issues please rebase on fresh `main`, I've just merged https://github.com/tonarino/innernet/pull/406

I'll wait a day or two with merging to give others (perhaps @bschwind / @mbernat / @goodhoko) also an opportunity to review.

And let me take the RPM build script question to https://github.com/tonarino/innernet/issues/203

mbernat (approved):
LGTM, thank you!

The `with_*` APIs all look tailored to file paths (rather than directory paths). Very strange there is no note about it in the docs and no warnings emitted, they just silently do the wrong thing.

